PR #7, opened 2019-08-16
report: styling improvements

tgorochowik:
The following things were improved:
* The cells that contain any test result now have a small dotted border.
  This helps to see what is what when the table header stops being
  visible.
* All the cells have now the same width (earlier it was aligned to the
  tool name).


mithro:
@tgorochowik Would have been good to attach a screenshot of the improvements to the pull request.

PR #9, opened 2019-08-16
Improve Makefile target generation performance

tgorochowik:
This significantly reduces the time of generating the targets given
there is a large number of tests.

For example for ~300 tests, the generation alone takes about 1 minute
and 30 seconds.

With this change it is almost instant.

### DIFF
--- a/tools/runner
+++ b/tools/runner
@@ -1,0 +1,42 @@
+#!/bin/env bash
+
+RUNNER="$1"
+TEST="$2"
+
+function parse_param() {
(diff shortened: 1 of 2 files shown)


tools/runner
```diff
@@ -0,0 +1,42 @@
+#!/bin/env bash
```
mithro: This should be a Python script I think....

tgorochowik: It was just extracted from the makefile, so making it a bash script was easier. However: https://github.com/SymbiFlow/sv-tests/pull/11

PR #11, opened 2019-08-19
tools: rewrite the runner wrapper in python

tgorochowik:
Also add some extra sanity check that the bash code did not have.


tgorochowik:
Pushed some minor PEP8 fixes

PR #13, opened 2019-08-20
[WIP]Add tree-sitter-verilog runner

piotr-binkowski:
I've created simple `tree-sitter-verilog` runner but in its current form it has two issues:

- It requires that `tree-sitter` executable and `grammar.js` are located relative to each other.

- It is significantly slower than other runners which is caused by the fact that it is always slower on its first run because it needs to generate parser library based on `grammar.js`. And due to using temporary directory for each run and because it looks for it in a directory relative to current directory, it must regenerate its library each time.


piotr-binkowski:
Performance issues should be fixed but now it is required that a `src` directory containing generated parser is located in the same directory as the `grammar.js` file

tgorochowik:
@piotr-binkowski Do you think it could be improved on top of #38 ?

We could now create a directory with all the needed wrappers in `tools/tree-sitter` for example (add the submodule inside together with the wrappers), build it just once and them just re-use it for each test?

piotr-binkowski:
If we ensure that `verilog.so` is built before running any tests then it should be possible achieve that using current runner script
But to get the `verilog.so` file we need to first parse a dummy file because the file seems to be created only when parsing and not during `generate` step

piotr-binkowski:
I've modified this runner to use `py-tree-sitter` and the `.so` file gets built in the setup step

This now depends on #67

PR #15, opened 2019-08-20
explicitly mark features with varied results

tgorochowik:
If one feature is tested by more than one test and the results of the tests vary, it should be clearly visible in the report.
    
This PR marks such features with orange.

Depends on #14 (and includes patches from there), feel free to comment but do not merge yet.


tgorochowik:
Rebased

PR #21, opened 2019-08-21
Add tests for the "Lexical conventions" chapter

tgorochowik:
Current status:

- [x] **5 Lexical conventions** - No separate tests, serves as an introduction for the next chapters
- [x] **5.1 General** - No separate tests, serves as an introduction for the next chapters
- [x] **5.2 Lexical tokens** - No separate tests, serves as an introduction for the next chapters
- [x] **5.3 White space** -  No separate tests, serves as an introduction for the next chapters (tested by basically every test, special white space treatment is verified by 5.9)
- [x] **5.4 Comments**
- [x] **5.5 Operators** - No separate tests, actual operators should be tested in chapter 11.3
- [X] **5.6 Identifiers, keywords, and system names**
- [X] **5.6.1 Escaped identifiers**
- [x] **5.6.2 Keywords** - these tests are all generated
- [X] **5.6.3 System tasks and system functions**
- [X] **5.6.4 Compiler directives** - note: the `include` directive is tested with `/dev/null`
- [x] **5.7 Numbers** - No separate tests, serves as an introduction for the next chapters
- [x] **5.7.1 Integer literal constants**
- [x] **5.7.2 Real literal constants**
- [x] **5.8 Time literals**
- [x] **5.9 String literals**
- [x] **5.9.1 Special characters in strings**
- [x] **5.10 Structure literals**
- [x] **5.11 Array literals**
- [x] **5.12 Attributes**
- [x] **5.13 Built-in methods**

Note:
The tests are heavily based on the examples from IEEE-1800-2017


tgorochowik:
I think all the chapters that have anything to test are now being tested.

Some notes:
 * a lot of tests - namely the test testing all the keywords - are generated. I am not sure if we shouldn't include some generators in the sources and dynamically generate them before running.

 * Another issue is that I think we should organize the tests differently. This PR adds tests for just a single chapter, and there is already lots of files which are just put directly in the tests directory. I think we need more sub-directories.

 * Odin seems to fail if the top module does not have any inputs/outputs. This makes sense for a simulator but does not for syntax checking (what we do here)

Including a screenshoot of the results of the tests for a quick reference:

tgorochowik:
* Removed the tests for 5.6.1, these are generated and included in #43 .
* Moved the tests to match the structure of the already merged PRs for other chapters.
* Rebased against master and removed the WIP label.

Please review.

PR #22, opened 2019-08-21
[WIP] Add Icarus Verilog simulation runner

piotr-binkowski:
Current Icarus runner only checks if the file gets successfully parsed but we should also check if it simulates successfully.
To achieve that test cases must be designed in such a way that if results match what was expected 
`:PASSED:` should be printed on simulator output.
In case the file tests only syntax it should contain an `initial` block with `$display(":PASSED:")` inside which would allow it to pass simulation test if it is parsed correctly.

### DIFF
--- a/runners/fake
+++ b/runners/fake
@@ -1,4 +1,5 @@
 #!/bin/bash
+#:type: syntax
 
 echo "I AM A FAKE RUNNER!"
 echo "MY INVOKATION ARGS: $@"
--- a/runners/icarus
+++ b/runners/icarus
@@ -1,3 +1,4 @@
 #!/bin/bash
+#:type: syntax
 
 iverilog -g2012 -o iverilog.out $1
--- a/runners/icarus-sim
+++ b/runners/icarus-sim
@@ -1,4 +1,5 @@
 #!/bin/bash
+#:type: simulation
 
 OUT="iverilog.out"
 
--- a/runners/icarus-sim
+++ b/runners/icarus-sim
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+OUT="iverilog.out"
+
+iverilog -g2012 -o $OUT $1
+vvp $OUT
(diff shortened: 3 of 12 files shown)


ldalek-antmicro:
I don't like it either. I think that test result should be determined by inspecting outputs of module after proper stimulation its inputs.

About additional flags. Maybe `steps` field that will specify what steps are executed on test. For example: `steps: syntax` would only run syntax checking but `steps: syntax, simulation` would run both. Of course tools like `yosys` would be skipped for simulation step.

piotr-binkowski:
I've added support for simulating only a selected subset of tests by adding tags to runners and tests,
but I'm still wondering how we should go about verifying the results.

I don't think we can follow the easiest way of directly comparing `.vcd` simulator output with a reference output because first of all we would require an external simulator that fully supports all SystemVerilog features that we intend to test. And moreover its `.vcd` output would need to be the indistinguishable from what tested simulators generate or we would need to manually fine-tune the reference `.vcd` files beforehand.

My idea to solve that was to create such test cases that after simulating would print out on its own if the test succeeded as that would allow us to not deal with `.vcd` files but maybe someone would come up with a better solution to this obstacle.

tgorochowik:
I think this approach is better.

I am thinking about the report now - if the tests/runners are now divided into groups, maybe the report should be somehow divided too?

We could add some global "tabs" on the top of the report: [syntax] [simulation]
Or maybe the reports should be completely separate html files (the report generator could take some parameters)
Open questions:
- will the input database be the same for various steps?

(please don't implement the above yet, I am just asking for opinions)

piotr-binkowski:
I think that using separate tabs or even files for each type would be taking this a bit too far, maybe we could create column groups and separate them from each other based on runner type.

With regards to shared/separate input database I think that as long as we test only the synthesizable subset of SystemVerilog each step could test the same features and use shared input database.

But if coverage by various types of runners would differ greatly then it might be good idea to make the results more readable by using separate tabs/files or even input databases.

Another option would be to automatically hide chapters without any results which would allow us to use the same input database without impacting readability.

PR #39, opened 2019-08-27
Add initial travis config

tgorochowik:
Initial travis configuration.


mithro:
Looks like travis is running successfully!

tgorochowik:
Yup, I guess it is a good start.

Two things are checked are the moment:
* PEP8 checks for pythons scripts in the `tools` directory
* Whether the `make report` target builds successfully 

@mithro Could you help with this by setting up github pages and adding a a secret variable with a deploy key/github token? (as per https://docs.travis-ci.com/user/deployment/pages/)

I guess the report deployment can be done separately so I removed the `[WIP]` label, but it seems I cannot setup the gh-pages myself.

PR #40, opened 2019-08-27
tests: add various tests

piotr-binkowski:
This adds various tests for chapters 5, 6, 7 and 8


piotr-binkowski:
All new tests should be now located inside `tests/generic/`

PR #47, opened 2019-08-30
Add sv2v runner

### LINKED ISSUE #45: Add sv2v to the set of tools to test
This looks like a promising converter that accepts System Verilog (and converts it to Verilog).

https://github.com/zachjs/sv2v

Might be good to also add to the set of tools we test.

piotr-binkowski:
Closes #45

### DIFF
--- a/runners/zachjs-sv2v
+++ b/runners/zachjs-sv2v
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+zachjs-sv2v $1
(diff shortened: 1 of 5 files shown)


mithro:
There are multiple projects call sv2v - Can you call this by the username + sv2v?

piotr-binkowski:
I've renamed the runner to `zachjs-sv2v` which should make it clearly distinguishable from any other `sv2v` projects

runners/zachjs-sv2v
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+zachjs-sv2v $1
```
tgorochowik: Will this not be problematic when we do #54 ?

I understand the need to include "zachjs" in the runner name, but I am not sure about the binary name.

PR #49, opened 2019-08-30
Add tests for chapter 7

ldalek-antmicro:
Add tests for chapter 7. It has missing tests for subsections 7.10.3 and 7.10.5.

### DIFF
--- a/tests/chapter-7/arrays/associative/alloc.sv
+++ b/tests/chapter-7/arrays/associative/alloc.sv
@@ -1,0 +1,17 @@
+/*
+:name: associative-arrays-allocating-elements
+:description: Test associative arrays elements allocation
+:should_fail: 0
+:tags: 7.8.7 7.8 7.9.1
+*/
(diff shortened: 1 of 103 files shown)


ldalek-antmicro:
All tests are generated from `basic7.sv` template.

Generator can be adapted to fit #43  and `basic7.sv` template can be moved to e.g. `conf/generators/template-basic7.sv`

ldalek-antmicro:
Pushed whole chapter-7 test suite. It includes tests with asserts for testing support of functions such as `sort` or `find`.

tgorochowik:
@hzeller the tests here implement the features that could be used for testing simulation.

I think there is no point delaying this PR as these tests will still work just fine for syntax checking, so should be included in the report as soon as possible.

We still have to decide if we want to go this way (and adjust the runners to eval the `:assert:` and `:re:` and then modify the remaining tests to also do that where applicable)

Issue #33, opened 2019-08-27
Capture all the version information about the tools


mithro:
Capture exact information about the version of all tools (IE Git revision etc).

PR #63, opened 2019-09-03
Add information about the versions of the tools

tgorochowik:
This PR changes the structure of the runners. Instead of a single
script file, there are now directories with single script files.

The reason is that the directory will be used as the displayed name of
the runner (e.g. slang), but the script name will be used to determine
where does the actual tool binary come from (e.g. driver for slang).

This is needed to be able to find out whether the tool that is tested
comes from a submodule or is provided externally.

The tool is now versioned in two ways:
 * using the appropriate `--version` switch for the tool
 * if it comes from the submodule, using the commit hash from git

If any of the above is impossible `n/a` is shown instead of the version string/commit hash.

Because there is no standard `--version` switch for all the tools, the
runners as the wrappers have to now support this universal switches:
 * `--version` - to call the internal version getting switch of the tool
 * `--test TEST` - to run the actual test

This PR adds that to all the currently existing runners.

The version can be viewed in the tooltip in the report. The tooltip is displayed when hovering the tool names with a mouse, for example:

Closes: #33

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,13 @@ $(OUT_DIR)/logs/$(1)/$(2).log: $(TESTS_DIR)/$(2)
 tests: $(OUT_DIR)/logs/$(1)/$(2).log
 endef
 
+define runner_version_gen
+$(OUT_DIR)/logs/$(1)/version:
+	./tools/runner --runner $(1) --version --out $(OUT_DIR)/logs/$(1)/version
+
+versions: $(OUT_DIR)/logs/$(1)/version
+endef
+
 define generator_gen
 generate-$(1):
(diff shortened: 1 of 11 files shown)


tgorochowik:
Added a WIP label as we may want to remove the git submodules and provide everything externally.

mithro:
Output looks good to me. Not reviewed the actual code changes.

tgorochowik:
Converted the code to python (the original implementation was still based on bash runners) and simplified it a little bit to only query the version from the tool (--version by default)

It now looks like this:


I guess it would make sense to add some info about the version from conda packages, but that can be done separately.

@piotr-binkowski would you mind taking a look?

PR #65, opened 2019-09-04
Manually force ccache usage

tgorochowik:
PR to test some travis ccache settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ cache:
   directories:
     - $HOME/.ccache
     - $HOME/.stack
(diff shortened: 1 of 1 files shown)


tgorochowik:
Using `ccache` properly and caching `.conan` reduced the time from [29 min 47 sec](https://travis-ci.com/SymbiFlow/sv-tests/jobs/231038470) to [17 min 9 sec](https://travis-ci.com/SymbiFlow/sv-tests/jobs/231049547).

It is still not perfect but it is a noticeable improvement. I think we should merge it.

Issue #60, opened 2019-09-03
Add swerv as a test


mithro:
https://github.com/westerndigitalcorporation/swerv_eh1

PR #67, opened 2019-09-05
runners: rewrite runners to use python

piotr-binkowski:
This converts old runner shell scripts into python classes which makes passing parameters to runners much easier.
On top of that this also adds support for 2 new metadata fields `:files:` and `:incdirs:` which allow to create tests using multiple files or requiring specifying include directories.
If any particular test case doesn't specify values for those fields then by default `:files:` points to test file itself and `:incdirs:` is empty

This could prove useful for supporting large test cases like #60


tools/Runner.py
```diff
@@ -0,0 +1,6 @@
+class Runner:
```
tgorochowik: We shouldn't have a `Runner.py` and a `runner` python script in the same directory. We need to organize  it better.

tools/runners/Yosys.py
```diff
@@ -0,0 +1,34 @@
...
+            for svf in params['files']:
+                f.write('read_verilog -sv' + inc + ' ' + svf + '\n')
+
+        proc = subprocess.Popen(cmd, cwd=tmp_dir,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.STDOUT)
+
+        log, _ = proc.communicate()
+
+        return (log.decode('utf-8'), proc.returncode)
```
tgorochowik: It seems that all of the runners share the last part (running the sub-process and returning the same stuff). Why is this not some common code if we have a common `Runner.py` anyway?

tools/Runner.py
piotr-binkowski: I've renamed `Runner.py` to `BaseRunner.py` to at least make filenames less confusing

.travis.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       install:
         - pip install -r conf/requirements.txt
       script:
-        - "flake8 $(file generators/* tools/* | sed -ne 's/:.*python.*//pI')"
+        - "flake8 $(file generators/* tools/* tools/runners/* | sed -ne 's/:.*python.*//pI')"
```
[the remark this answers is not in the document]
piotr-binkowski: I've added another line and `flake8` will now also check `tools/*.py` and `tools/runners/*.py` files

tools/runner
```diff
@@ -27,6 +34,16 @@ ch = logging.StreamHandler()
...
 
+runners = {}
+
+runners['Fake'] = Fake()
+runners['Slang'] = Slang()
+runners['Icarus'] = Icarus()
+runners['Yosys'] = Yosys()
+runners['Sv2v_zachjs'] = Sv2v_zachjs()
+runners['Odin'] = Odin()
+runners['Verilator'] = Verilator()
```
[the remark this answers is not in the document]
piotr-binkowski: Runner modules and classes are now handled using `import_module` and `getattr` which replaced direct instantiation

tools/runners/Yosys.py
piotr-binkowski: This common part is now inside `BaseRunner` and is shared by almost all runners. `Fake` runner overrides this as it must set the return code based on what is expected by the test case

tools/runners/Icarus.py
```diff
@@ -0,0 +1,15 @@
...
+        super().__init__("icarus")
+
+    def run_cb(self, tmp_dir, params):
+        self.cmd = ['iverilog', '-g2012', '-o iverilog.out']
+
+        for incdir in params['incdirs']:
+            self.cmd.append('-I' + incdir)
+
+        for f in params['files']:
+            self.cmd.append(f)
```
tgorochowik: Why not just `cmd += params['files']` instead of the loop?

tools/runners/Odin.py
```diff
@@ -0,0 +1,16 @@
...
+class Odin(BaseRunner):
+    def __init__(self):
+        super().__init__("odin")
+
+    def run_cb(self, tmp_dir, params):
+
+        self.cmd = ['odin_II', '--permissive', '-o odin.blif', '-V']
+
+        # odin doesn't seem to support include directories
+        # and thus only list of files is provided to it
```
tgorochowik: What does that mean? What are the implications of that?

piotr-binkowski:
I've added support for third meta parameter `:top_module:` which is a must in cases where multiple source files are used

tgorochowik:
@piotr-binkowski the `:top_module:` change looks good. Please look at the minor stuff I pointed out in my previous review and I think we're good to go!

tools/runners/Odin.py
piotr-binkowski: ODIN II tool doesn't list any way to provide list of include directories in its help message (`odin_II -h`) and this will be a big issue for tests that depend on that, ie. `SweRV` and `Ibex` cores use include directories for their headers

tools/BaseRunner.py
```diff
@@ -0,0 +1,17 @@
+import subprocess
+
+
+class BaseRunner:
+    def __init__(self, name):
+        self.name = name
+
+    def run(self, tmp_dir, params):
```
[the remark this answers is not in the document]
piotr-binkowski: I've added some basic information about runners to `BaseRunner.py`

tools/runners/Icarus.py
piotr-binkowski: This should be now fixed in all the runners

.travis.yml
```diff
@@ -24,6 +24,7 @@ jobs:
         - pip install -r conf/requirements.txt
       script:
         - "flake8 $(file generators/* tools/* | sed -ne 's/:.*python.*//pI')"
+        - "flake8 $(file tools/*.py tools/runners/*.py)"
```
tgorochowik: I think we cannot do it like this after all. It needs to be a part of the same command, otherwise if the first check fails, the second is not run.

piotr-binkowski: `flake8` now checks all files in a single command

mithro:
@hzeller - Python scripts are much easier to understand than shell once they get over 5 lines. Also error handling in shell scripts is almost impossible.

hzeller:
@mithro My point is, that the runners should be as simple as possible, they are merely adapters to the binaries in question. They should never really be longer than 5 lines. We should keep the complexity in the places that invokes stuff.

Current Python wrappers do a somewhat awkward way to set environment variables and invoke the binary ... which arguably is _much_ more readable in a shel-lscript, as this is its bread and butter.

If there is any pre-processing, we should provide that in the Python script that invokes that adapter.

As an intermediate step, Python is fine, but as soon as we want to advertise it to people to use, shell-scripts will sell better.

PR #68, opened 2019-09-05
Add SweRV as a test

### LINKED ISSUE #60: Add swerv as a test

piotr-binkowski:
This adds SweRV core as a test with its separate tag `swerv`

Closes #60 
Depends #67

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "tools/zachjs-sv2v"]
 	path = tools/zachjs-sv2v
 	url = https://github.com/zachjs/sv2v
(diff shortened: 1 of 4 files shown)


piotr-binkowski:
Renaming `cores` to something more generic like `ext` or even `external` is a good idea. Especially if those imported tests will be handled in a similar way

mithro:
The convention we use in SymbiFlow is `third_party` see https://github.com/SymbiFlow/symbiflow-arch-defs/tree/master/third_party

PR #70, opened 2019-09-10
Move third-party submodules to third_party directory

piotr-binkowski:
As mentioned in #68, SymbiFlow uses `third_party` directory for adding this kind of submodules and we probably should do the same

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,21 +1,21 @@
-[submodule "tools/runners/odin_ii"]
-	path = tools/odin_ii
+[submodule "third_party/tools/odin_ii"]
+	path = third_party/tools/odin_ii
 	url = https://github.com/verilog-to-routing/vtr-verilog-to-routing.git
-[submodule "tools/runners/icarus"]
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -16,6 +16,6 @@
 [submodule "tools/zachjs-sv2v"]
```
mithro: FYI - `tools/zachjs-sv2v` should probably be `third_party/tools/zachjs-sv2v`

.gitmodules
```diff
@@ -16,6 +16,6 @@
...
 	path = tools/zachjs-sv2v
 	url = https://github.com/zachjs/sv2v
-[submodule "external/swerv"]
```
mithro: With the above comment about `third_party/tools/zachjs-sv2v` this should probably end up as something like `third_party/cores/swerv` or `third_party/tests/swerv` or similar?

piotr-binkowski:
I've moved all the submodules to their appropriate locations inside `third_party` directory

PR #72, opened 2019-09-10
Import tests from ivtest

### LINKED ISSUE #57: Import tests from ivtest
https://github.com/steveicarus/ivtest

piotr-binkowski:
This adds generator that imports most of the tests in `ivtest`

In it's current form it has several issues:
- It imports most of the tests but some should be omitted(VHDL integration, `Verilog-AMS`)
- Our `Makefile` can't handle that much tests in the `info` target and `@echo -e "Found the following tests:$(subst $(space),"\\n \* ", $(TESTS))\n"` had to be removed
- It makes the report page substantially slower, ie. switching between results of individual tests takes seconds
- `report.html` gets as big as 40MB which is too big to serve as a single static webpage and probably we should somehow load the results in a more dynamic way or at least store the logs in separate files

Closes #57 
Updated version depends on #117

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -55,3 +55,6 @@
 [submodule "third_party/cores/rsd"]
 	path = third_party/cores/rsd
 	url = https://github.com/rsd-devel/rsd
+[submodule "third_party/tests/ivtest"]
+	path = third_party/tests/ivtest
+	url = https://github.com/steveicarus/ivtest.git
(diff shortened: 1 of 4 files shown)


piotr-binkowski:
- I've fixed the issue with tests that should be omitted by adding a blacklist of list files
- Issues with reports were fixed by using #117

This now depends on #117

mithro:
@piotr-binkowski ping? Can you rebase this onto the latest master and resubmit?

piotr-binkowski:
I've rebased it, lets see if it passes the CI

PR #79, opened 2019-09-12
Add tests for chapter "Data types"

piotr-binkowski:
Depends #99


piotr-binkowski:
I think that this is a good idea to use a common generator script and config files specific to each case, I will create a separate PR for this feature

tgorochowik:
@piotr-binkowski the dependencies have been merged. Please rebase

PR #99, opened 2019-09-16
Merge similar generators into single script with config files

piotr-binkowski:
This replaces some of the generator scripts with a common one and a set of config files to reduce code duplication


conf/generators/assignment.json
```diff
@@ -0,0 +1,34 @@
...
+		":tags: 11.4.1",
+		"*/",
+		"module top();",
+		"int a = 12;",
+		"int b = 5;",
+		"initial begin",
+		"    a {0} b;",
+		"end",
+		"endmodule"
+	],
```
tgorochowik: Why does this have to be an array?

conf/generators/assignment.json
```diff
@@ -0,0 +1,34 @@
+{
+	"name": "assignment_op",
+	"fname": "11.4.1--{1}_assignment.sv",
+	"templ": [
```
tgorochowik: Since it is becoming a kind of API for tests generators, maybe we could use the full `template` word here?

generators/generator_json
```diff
@@ -0,0 +1,36 @@
...
+except KeyError:
+    print("Export the TESTS_DIR and CONF_DIR variables first")
+    sys.exit(1)
+
+name = ''
+fname = ''
+templ = ''
+vals = []
+
+for cfg in glob.glob(os.path.join(conf_dir, 'generators', '*.json')):
```
tgorochowik: maybe we could use a subdir for all the inputs for this generator? I think it is possible that other generators may want to use jsons for configuration too.

generators/keywords
```diff
@@ -28,13 +28,15 @@ except IndexError:
...
+
 conf_file = os.path.join(conf_dir, "generators", "sv-keywords.list")
 with open(conf_file, "r") as f:
     for x in f:
         x = x[:-1]
-        test_file = os.path.join(tests_dir,
-                                 "generated",
-                                 tests_subdir,
-                                 "5.6.2--keyword-" + x + ".sv")
+        test_file = os.path.join(test_dir, "5.6.2--keyword-" + x + ".sv")
```
tgorochowik: Any reason why the keyword tests cannot use the same generator as the other tets with a simple template?

conf/generators/assignment.json
piotr-binkowski: Good idea, should `fname` be expanded to `file_name` too?

generators/generator_json
piotr-binkowski: I agree, a subdir should be used to make this more future proof

conf/generators/assignment.json
piotr-binkowski: Is it possible to use multiline variables in JSON?
When I tried to use a single string taking multiple lines it failed parsing in the generator script

generators/keywords
piotr-binkowski: I don't think that there are any, I will update the `keywords` generator to use this new approach

PR #100, opened 2019-09-17
Update 5.6.4--compiler-directives-unconnected-drive.sv

alainmarcel:
This is the proper use of this directive


tgorochowik:
@alainmarcel Thank you for your contributions!

There is one minor issue with the DCO in your commit. Symbiflow requires you to use a full name in the signed-off line.

Could you please `--amend` the commit so the line contains your full name? It should look like this
```
Signed-off-by: Alain Dargelas <alainmarcel@yahoo.com>
```

To make it automatic you should configure your git client like this:
```
git config --global user.name "FIRST_NAME LAST_NAME"
```

The signed-off line should be ok out of the box then.

Thanks!

PR #104, opened 2019-09-17
Reduce runner console log verbosity

### LINKED ISSUE #81: Runners should be less verbose
Probably from the early implementation, there are still some debugging messages in the runner code that output how a test was successful.

 *  [116: logger.info("Running {}/{} ...](https://github.com/SymbiFlow/sv-tests/blob/master/tools/runner#L116)
 *  [129:logger.info("PASS: {}/{}".format(args.runner,](https://github.com/SymbiFlow/sv-tests/blob/master/tools/runner#L129)
 *  [131:logger.warning(](https://github.com/SymbiFlow/sv-tests/blob/master/tools/runner#L131)

These are merely FYI outputs of expected results, so should not clutter the output here: they that end up in the log-files anyway, to be reported by the `sv-report`. So I'd either remove them , make them logger.debug() or set the [default log-level](https://github.com/SymbiFlow/sv-tests/blob/master/tools/runner#L24) to ERROR (probably the best choice).

Problem would be that these will drown out actual problems, such as the Exception reported here:
https://github.com/SymbiFlow/sv-tests/blob/master/tools/runner#L140

piotr-binkowski:
Currently `tools/runner` was configured for `DEBUG` logging level but as suggested in #81 it should be reduced

Closes #81

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ runners:
 # $(2) - test
 define runner_gen
 $(OUT_DIR)/logs/$(1)/$(2).log: $(TESTS_DIR)/$(2)
-	@./tools/runner --runner $(1) --test $(2) --out $(OUT_DIR)/logs/$(1)/$(2).log
+	@./tools/runner --runner $(1) --test $(2) --out $(OUT_DIR)/logs/$(1)/$(2).log --quiet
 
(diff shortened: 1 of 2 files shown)


piotr-binkowski:
I've added `--quiet` switch to show only error messages which is now used in the `Makefile`, old logging level is accessible by omitting this switch

PR #105, opened 2019-09-17
Check if tool is installed before using it

### LINKED ISSUE #83: Execution of tools should depend on availability of these tools.
Right now, the `RUNNERS` are filled by looking at the corresponding `tools/runners/*.py` scripts.

However, this requires that all the tools mentioned there have been checked out and compiled before into `out/runners`. If any of these has not been checked out or compiled, the whole test-suite run will fail.

The typical use-case in the future will be, that the users of this test-suite will only check out the tools they are interested in, and want to have a report created from these.

So we should look that we make the dependency what tools are run in the test-suite dependent on the existence of the necessary binaries.

This could be either done with make-file means (but then this would mean that the Makefile needs to have the knowledge what tools are there), or by having something like a probing `--can_run` flag on the  scripts in the `tools/runners/` directory: only the ones that return `exit(0)` are included in the report generation.

piotr-binkowski:
Currently all tools are used without checking if a particular tool is installed, in #83 it was suggested to test if tools are present before using them and this adds support for checking that.
If tool is not present then testing it will be skipped and it won't be included in the report.

Closes #83

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,8 @@ export TESTS_DIR
 export RUNNERS_DIR
 export GENERATORS_DIR
 
+USE_ALL_RUNNERS?=0
+
 include tools/runners.mk
 
 .PHONY: clean init info tests generate-tests report
@@ -47,9 +49,10 @@ generate-$(1):
 generate-tests: generate-$(1)
 endef
 
-RUNNERS := $(wildcard $(RUNNERS_DIR)/*.py)
-RUNNERS := $(RUNNERS:$(RUNNERS_DIR)/%=%)
-RUNNERS := $(basename $(RUNNERS))
+RUNNERS_FOUND := $(wildcard $(RUNNERS_DIR)/*.py)
+RUNNERS_FOUND := $(RUNNERS_FOUND:$(RUNNERS_DIR)/%=%)
+RUNNERS_FOUND := $(basename $(RUNNERS_FOUND))
+RUNNERS := $(shell OUT_DIR=$(OUT_DIR) ./tools/check-runners $(RUNNERS_FOUND))
 TESTS := $(shell find $(TESTS_DIR)/ -type f -iname *.sv)
 TESTS := $(TESTS:$(TESTS_DIR)/%=%)
 GENERATORS := $(wildcard $(GENERATORS_DIR)/*)
@@ -58,6 +61,11 @@ GENERATORS := $(GENERATORS:$(GENERATORS_DIR)/%=%)
 space := $(subst ,, )
 
 info:
+ifneq ($(USE_ALL_RUNNERS), 0)
+ifneq ($(RUNNERS), $(RUNNERS_FOUND))
+	$(error Some runners are missing)
+endif
+endif
 	@echo -e "Found the following runners:$(subst $(space),"\\n \* ", $(RUNNERS))\n"
 	@echo -e "Found the following tests:$(subst $(space),"\\n \* ", $(TESTS))\n"
 
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -1,4 +1,5 @@
 import subprocess
+import shutil
 
 
 class BaseRunner:
@@ -13,8 +14,15 @@ class BaseRunner:
     to be detected and launched by the Makefile.
     """
 
-    def __init__(self, name):
+    def __init__(self, name, executable=None):
+        """Base runner class constructor
+        Arguments:
+        name -- runner name.
+        executable -- name of an executable used by the particular runner
+        can be omitted if default can_run method isn't used.
+        """
         self.name = name
+        self.executable = executable
 
     def run(self, tmp_dir, params):
         """Run the provided test case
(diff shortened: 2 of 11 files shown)


tools/runners/Icarus.py
```diff
@@ -4,9 +4,10 @@
 class Icarus(BaseRunner):
     def __init__(self):
         super().__init__("icarus")
+        self.ex = 'iverilog'
```
tgorochowik: Can we rename `ex` to something self-explanatory?

tools/BaseRunner.py
```diff
@@ -34,3 +35,11 @@ def run(self, tmp_dir, params):
         log, _ = proc.communicate()
 
         return (log.decode('utf-8'), proc.returncode)
+
+##
+# This method is called by the main runner script (tools/runner) as a sanity
+# check to verify that tool is properly installed.
+# If False is returned then the test is skipped.
+# @return True when tool is installed and can be used, False otherwise.
```
tgorochowik: Why are the comments not a proper python doc strings?

Please convert the comments here (and everywhere) using the proper conventions (see https://www.python.org/dev/peps/pep-0257/ )

tools/runner
```diff
@@ -54,6 +54,9 @@ new_path = [os.path.abspath(dirs['out'] + "/runners/bin/"), os.environ['PATH']]
 
 os.environ['PATH'] = ":".join(new_path)
 
+if not runner_obj.can_run():
+    sys.exit(0)
```
tgorochowik: How does this resolve #83? 

To actually solve #83 I think that we should exclude the runner from the report completely, not "fake-pass" all the tests.

Additionally I don't think it should be quiet. We should at least show a warning about skipping a runner.

Thirdly I think there still should be a way to force-require all runners (could be a Makefile flag). We don't want to exclude anything in CI by accident.

piotr-binkowski: By doing a "fake-pass" at this stage no log file is generated and thus the column for a particular tool won't be included in the report.

Detecting available tools could be moved to main Makefile and we could handle the warnings and hard requirement for all runners there

tools/runners/Icarus.py
piotr-binkowski: `executable`?

tools/runner
tgorochowik: That makes sense. But still - at least a part of it should be moved to makefile (to be able to error on this in CI).

I am also wondering if we could somehow not invoke the runners that fail on can run (am I right to think that this will be executed for each test anyway? We should be able to check it just once).

tools/runners/Icarus.py
tgorochowik: +1

A comment about this in the main class would be nice too

tgorochowik: Maybe the executable could also be a kwarg in the main class constructor?

tgorochowik: So you could just call `super().__init__("icarus", "iverilog")` etc

piotr-binkowski:
I've fixed the minor issues and moved the tool checks to makefile 
It is now possible to provide flag `ALL_RUNNERS=1` to makefile that forces it to fail if any of the tools is missing

tools/BaseRunner.py
```diff
@@ -34,3 +44,13 @@ def run(self, tmp_dir, params):
...
+
+    def can_run(self):
+        """Check if runner can be used
+        This method is called by the main runner script (tools/runner) as
+        a sanity check to verify that tool used by the runner is properly
+        installed.
+
+        Returns True when tool is installed and can be used, False otherwise.
+        """
+        return True if shutil.which(self.executable) is not None else False
```
hzeller: Can't this be formulated more readably as
```
  return shutil.which(self.executable) is not None
```
We return a boolean, so not need to convert a boolean to a boolean :)

piotr-binkowski: Thanks for pointing that out, it looks better this way

Makefile
```diff
@@ -47,9 +49,10 @@ generate-$(1):
 generate-tests: generate-$(1)
 endef
 
-RUNNERS := $(wildcard $(RUNNERS_DIR)/*.py)
-RUNNERS := $(RUNNERS:$(RUNNERS_DIR)/%=%)
-RUNNERS := $(basename $(RUNNERS))
+RUNNERS_FOUND := $(wildcard $(RUNNERS_DIR)/*.py)
+RUNNERS_FOUND := $(RUNNERS_FOUND:$(RUNNERS_DIR)/%=%)
+RUNNERS_FOUND := $(basename $(RUNNERS_FOUND))
```
tgorochowik: Can we use this variable when calling the `runner_gen` macro? Otherwise we get lots of "empty runs" I mentioned in my previous comment.

piotr-binkowski: Maybe the variable names are misleading but `RUNNERS` is the variable containing runners that passed the checks and `RUNNERS_FOUND` contains all runners which are present in `tools/runners`.

tgorochowik: Ok, great, thanks!

PR #106, opened 2019-09-17
Add tag description to report page and left align tag name

### LINKED ISSUE #75: Left align the test name
Would make the following line up nicely;
```
5
5.1
5.1.1
5.1.2
5.1.3
5.2
5.2.1
5.2.2
```

Could maybe expand to a tree view?

### LINKED ISSUE #84: conf/lrm.conf should be used for descriptive annotations in report.html
The config file that maps sections to descriptions, but right now, this file is not used much.

The `sv-report` uses it to create a list of potential test case classifications, but we should possibly also use the information there to annotate the output in the reports. For instance having 11.4.7 annotated as `11.4.7  Logical operators` is probably very helpful.

piotr-binkowski:
This adds column with tag description to the report page
Screenshot_20190917_143048
Closes #84
Closes #75

### DIFF
--- a/conf/report-template.html
+++ b/conf/report-template.html
@@ -102,8 +102,10 @@
         $('#report_table').DataTable( {
           paging: false,
           fixedHeader: true,
-          "columnDefs": [ { "type": "sv-id", targets: 0 } ],
+          "order": [[ 1, "asc" ]],
+          "columnDefs": [ { "type": "sv-id", targets: 1 } ],
           "columns": [
+            null,
             null,
             {% for tool in report %}
             { "orderDataType": "test-status" },
@@ -119,8 +121,22 @@
         width: auto;
         table-layout: fixed;
       }
-      table.dataTable tbody tr th {
-        width: calc(100% / {{ report.keys()|length + 1}} );
+      table.dataTable tbody tr th#report_table_result {
+        width: calc(80% / {{ report.keys()|length}} );
+      }
+
+      table.dataTable tbody tr th#report_table_tag {
+        text-align: left;
+        width: 50px;
+      }
+
+      table.dataTable tbody tr th#report_table_info {
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        text-align: left;
+        overflow: hidden;
+        max-width: 150px;
+        width: 150px;
       }
 
       .ui-widget-shadow {
@@ -339,6 +355,7 @@
       <thead>
         <tr>
           <th>  </th>
+          <th>  </th>
           {% for tool in report %}
           <th> {{ tool.lower() }} </th>
           {% endfor %}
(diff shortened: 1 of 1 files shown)


conf/report-template.html
```diff
@@ -119,8 +120,16 @@
         width: auto;
         table-layout: fixed;
       }
-      table.dataTable tbody tr th {
-        width: calc(100% / {{ report.keys()|length + 1}} );
+      table.dataTable tbody tr th#result {
+        width: calc(80% / {{ report.keys()|length}} );
```
tgorochowik: Why 80%?

conf/report-template.html
```diff
@@ -338,6 +347,7 @@
     <table id="report_table">
       <thead>
         <tr>
+          <th>  </th>
```
tgorochowik: Maybe we should add the headers here? It looks weird if the first two columns don't have any headers

conf/report-template.html
```diff
@@ -346,9 +356,10 @@
       </thead>
         {% for tag, info in database.items() %}
         <tr>
-          <th title="{{info}}"> {{ tag }} </th>
+          <th id="tag"> {{ tag }} </th>
```
tgorochowik: Could you use longer ids to avoid possible conflicts?

conf/report-template.html
```diff
@@ -346,9 +356,10 @@
...
+          <th id="info"> {{ info }} </th>
           {% for tool, tooldata in report.items() %}
-          <th class="{{ tooldata["tags"][tag]["status"] }}
+          <th id="result" class="{{ tooldata["tags"][tag]["status"] }}
```
tgorochowik: Maybe we could add `text-overflow: ellipsis;` to this? It looks weird if the test boxes do not all have the same height.

tgorochowik: Btw assuming we do add this, please restore the title (tooltip). We could actually add this to both columns (id + description)

tgorochowik: BTW. This comment should of course be under the ``{{ info }}`` change.

conf/report-template.html
piotr-binkowski: There is no reason for this particular value but when set to 100% the results table fills the whole width which wasn't happening previously. Should we use the full width?

piotr-binkowski:
After adding requested changes and left aligning the tag column as was suggested in #75, it looks like this now
Screenshot_20190917_152718

piotr-binkowski:
After some additional minor fixes it now looks like this
Screenshot_20190917_154218

hzeller:
LGTM
I like the info column to the left of the tag, but if it is was right aligned, it might make it more readable in some cases. But no strong preference, I am happy with this overall.

PR #111, opened 2019-09-18
Wrong assignment_pattern form when using "default"

alainmarcel:
Wrong assignment_pattern form when using "default" 
No semi-column allowed after begin or end keywords


mithro:
@alainmarcel - Could you update the pull request title to be more informative?

PR #117, opened 2019-09-19
Dynamic log loading

### LINKED ISSUE #115: Use externally stored logs in report page
As the test suite gets bigger, storing all the logs in the `report.html` file starts to create issues with the resulting file size (currently around 12MB) and performance.
Maybe loading the logs from external storage in a dynamic way could help with that.

piotr-binkowski:
Storing logs in the `report.html` file starts to become troublesome as explained in #115
This modifies the report page to only provide relative links to log files instead of displaying them in a box above the list of tests.
With this PR, clicking on a particular test should open a new browser tab with log file for that testcase

Closes #115


hzeller:
mmh, so instead of a separate tab, we ideally should see the logs in a frame or so similar to before. That way, we can inspect the log-files quickly.

I'll hack up a quick sample.

hzeller:
Easiest sample is with code, so made #118 for discussion.

piotr-binkowski:
Thanks to sample from #118 I've modified the page to look like this:
Screenshot_20190920_134007

piotr-binkowski:
I've updated the code to support `gh-pages` deployment while still supporting links to sources by copying tests to `out` directory and using relative paths

Issue #85, opened 2019-09-12
Available tags in tests should drive the report, not tags available in lrm.conf


hzeller:
**What happens now**
Currently, all tags found in the test-cases are matched against availability in the [lrm.conf](https://github.com/SymbiFlow/sv-tests/blob/master/conf/lrm.conf) and "Tag not present in the database"-warned when it is not, and the test is removed.
The keys in lrm.conf are then driving the lines that are reported.

This means, that someone who adds a test-case with a new tag also needs to be aware of the lrm.conf before a report can be generated. Given that the lrm.conf is mostly just helpful to provide descriptive annotations (after #84 is implemented), and potentially help define an order, this provides a larger hurdle to adding new test-cases that don't necessarily fit the tags we have available now.

**What would be desirable instead**
To simplify adding new tests, this should be the other way around: The tests declare tags, and `sv-report` will just collect all tags it has found in a map of sorts. We then output these and use the lrm.conf entries to define an order declared there (so that we keep the nice order of chapters which are otherwise  lexicographically mangled).

Tags that are in the database, but don't show up in any tests, should not be printed (this will avoid the gray lines we have now for all the tags that are not available yet).

Tags that are mentioned in the tests, but not in the database, are appended lexicographically at the end of the report.

(sv-report can still have some diagnostic output about keys not found in the database with `-v`).

Our goal is that the ultimate source of truth is the metadata in the *.sv tests themself: they should be self-contained and drive what is reported. Adding a new test should be as simple as adding a single file with its meta-data.

The lrm.conf can help present additional information by expanding tags to descriptive annotations (#84) but its content should not prevent adding new tests.

Issue #80, opened 2019-09-12
Makefiles: Avoid silent actions if possible


hzeller:
The makefiles currently have a lot of silent actions right now, which makes it harder understand what is happening and to debug what is going wrong when things fail. Often there is an `@echo` that describes what the following command does. We should let the commands speak for themself.

Let's avoid these to have a simpler and easier to use build-system. Here is an example on a subset of Makefile rules; instead of this:

```Makefile
clean:
	@echo -e "Removing $(OUT_DIR)"
	@rm -rf $(OUT_DIR)
	@echo -e "Removing $(TESTS_DIR)/generated/"
	@rm -rf $(TESTS_DIR)/generated/

report: init info tests
	@echo -e "\nGenerating report"
	@./tools/sv-report
	@echo -e "\nDONE!"

$(OUT_DIR)/logs/$(1)/$(2).log: $(TESTS_DIR)/$(2)
	@mkdir -p $(OUT_DIR)/logs/$(1)/$(dir $(2))
	@./tools/runner --runner $(1) --test $(2)
```

The following will be easier to read, debug and maintain:

```Makefile
clean:
	rm -rf $(OUT_DIR)
	rm -rf $(TESTS_DIR)/generated/

report: init info tests
	./tools/sv-report

$(OUT_DIR)/logs/$(1)/$(2).log: $(TESTS_DIR)/$(2)
	mkdir -p $(OUT_DIR)/logs/$(1)/$(dir $(2))
	./tools/runner --runner $(1) --test $(2)
```

(Possibly the only place where a silent action is useful is for `@echo` itself.)

Issue #35, opened 2019-08-27
Parse the error message


mithro:
It would be good to parse the error message and check that it is what is expected.

Nic30:
But how to make it compatible with all tools?

mithro:
@hzeller had ideas. I think he has doing this for https://github.com/zachjs/sv2v/pull/38

hzeller:
Idea is mostly to send patches to fix the tools. I did that for yosys last year, now currently sv2v, others to follow.

Nic30:
I have bad feeling that in some cases different tools can recognize different errors and both will be correct at the same time. Also errors in preprocessor may have very different formats.
However any standardisation of error messages would be great. I will check the commit in tools time to time so I know what is happening.

hzeller:
At this first step, we're concerned about the formatting of error messages to have the standard format of

```
<filename>:<line>:(<column>:)? (Error:|Warning:)? <msg>
```
This helps that common editors (such as emacs or vi) can easily step through error messages.
Some tools have more a 'free-form' version of this format.

Certainly, the tools will report the same problem in different ways or trip over problems in different ways, or have more information to be more verbose about things.
So I see less chance to always report exactly the same error - it will probably always somewhat of a difference you'd expect between different tools (think of `gcc` vs. `clang`)

With the line-number, we could at least narrow down the place where we expect the problem.

For the longer term of this scope, we should maybe provide some regular expression in the metadata of the tests that determine if error messages contain expected keywords; maybe metadata that can be specified for a particular tool.

hzeller:
As a follow-up, in the last year I sent a some patches to various tools so that we have a consistent errors/warnings oiutput in the common `<filename>:<line>(:col)?:` format, which allowed to provide the clickable messages that directly jump to the file in question.

Typically, we expect no error in the tests, so expectation is easy to formulate :)

For the ones that we actually *do* expect to fail, I think it will be hard to find an expected error with a robust regular expression. But at least we have #449 and #688 that adds a human readable expectation of why a test fails, which comes closest to the spirit of this bug.

I'd say, this bug is ready  to be closed ?

PR #118, opened 2019-09-20
Make test results easily explorable in one-page result.

### LINKED ISSUE #115: Use externally stored logs in report page

hzeller:
* Present all relevant information to explore tests on one page
   with multiple frames: Overview table with aggregates, list
   that shows all the tests and their outcomes, logfiles, sources.
 * Output is simple static HTML for fast rendering in browser.
 * Logfiles are HTML converted to contain color coding and links to
   source files.
 * Source files can be directly inspected by clicking on the links mentioning
   them.
 * Redact irrelevant information in logfiles (e.g. absolute paths where
   test suite runs)
 * Does not (yet) use templates (The template was using JavaScript
   to provide sorting functions which are not really used in practice.)
 * This means: everything relevant to create reports are in a single binary.

Fixes #115 (link external logfiles)
Also Fixes
  * #85 (use available logs, not lrm to show)
  * #80 (avoid silent actions)

Partially adresses #35 (now files mentioned in error messages are linked)

 TODO:
  * Links towards sources are currently leaving out/ structure:
    Source files would ideally also be re-created in the out/ directory
    with line-number <a name> tags and possibly simple syntax highlight.
  * Order of tests in the middle-column is in the sequence they were
    found on disk. Could be sorted.
  * Some parameters should be commandline flags.


hzeller:
This is a quick hack mostly for discussion at this point, roughly how I imagine an report output would be beneficial for the target group of the sv-tests-suite. It focuses on easy exploration and drill-down of the issues at hand (I didn't implement any column sorting or other 'neat' things, just enough to demo what I think is useful for the task).

sv-tests-report

For simplicity (for me), I wrote it in C++, it needs the following dependencies:

```
sudo apt-get install libboost-filesystem-dev libboost-regex-dev libboost-system-dev
```

(This is for demo of a suggested output report; It's ok to keep Python for now, though longer term, I'd really like to see the report-generator written in C++ for ease of maintenance).

piotr-binkowski:
The suggested page layout looks nice, based on it I've modified our current report page to look like this:
Screenshot_20190920_134007
Logs and sources are dynamically loaded using iframes but I still need to replace absolute paths to sources with relative ones and load the lists of tests using iframe

hzeller:
(Closing, as this was just for brainstorming)

PR #129, opened 2019-09-23
Import tests from hdlConvertor

piotr-binkowski:
Depends on #117

Slang seems to have issues with those tests on my PC, it uses up all the available RAM and swap on some of the cases, for example `tests/sv_test/std2017/p550.sv`. I wonder how this will behave in travis CI.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "third_party/cores/fx68k"]
 	path = third_party/cores/fx68k
 	url = https://github.com/ijor/fx68k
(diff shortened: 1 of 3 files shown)


piotr-binkowski:
It looks like Travis runs this without any big issues, only thing required was to modify the makefile to prevent Travis from failing after 10min of console inactivity.

This makefile change means that this now fixes #80

piotr-binkowski:
Makefile changes are now merged into master and now this PR only adds tests from hdlConvertor

PR #131, opened 2019-09-23
Add support for dynamic tag discovery

### LINKED ISSUE #85: Available tags in tests should drive the report, not tags available in lrm.conf

piotr-binkowski:
Currently all tags must be defined in `lrm.conf` file, with this adding entry there is only required to change the tag description from the default empty one.
Additionally any tags without tests are omitted from the report.

Closes #85

### DIFF
--- a/tools/sv-report
+++ b/tools/sv-report
@@ -168,6 +168,9 @@ for r in [os.path.dirname(r) for r in glob(args.logs + "/*/")]:
                     tags[tag]["status"].append(status)
                 except KeyError:
                     logger.warning("Tag not present in the database: " + tag)
+                    database[tag] = ''
+                    tags[tag] = {}
+                    tags[tag]["status"] = [status]
                     continue
 
         for tag in tags:
--- a/tools/sv-report
+++ b/tools/sv-report
@@ -54,6 +54,8 @@ elif args.verbose:
 else:
     logger.setLevel(logging.INFO)
 
+tag_usage = {}
+
 # generate input database first
 database = {}
 try:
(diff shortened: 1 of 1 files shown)


hzeller:
LGTM

PR #132, opened 2019-09-23
Import tests from billswartz7/utd-SystemVerilog

### LINKED ISSUE #55: Import tests from billswartz7/utd-SystemVerilog repostiory
https://github.com/billswartz7/utd-SystemVerilog/tree/master/tests/verilog

piotr-binkowski:
Closes #55
Depends #146


piotr-binkowski:
It appears that `utd-SystemVerilog` uses a submodule that is not publicly available and this makes the CI fail.

Only way I can think of that allows us to still include this repo is to handle submodule cloning manually, unless Travis offers some kind of submodule blacklist that would prevent it from trying to download that particular one.

mithro:
Let's just include them via copying in.

PR #139, opened 2019-09-23
[Obsolete, see #147] Not using silent rules; improve runner compilation parallelism

### LINKED ISSUE #80: Makefiles: Avoid silent actions if possible

hzeller:
* The silent rules made it hard to follow what is going on.
 * If we use `$(MAKE)` directly in a rule, it is able to inherit
   other configuration from the inherited make, such as -j; this
   makes compilation much faster in these cases.
  * Instead of `cp`, use `install` to install binaries
     * It prevents issues when the binary is in use at the time
          for some reason.
      * With -D, we can make sure that directories are created, so
          no need for our own `mkdir -p`
    
Fixes #80

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,10 +19,8 @@ include tools/runners.mk
 .PHONY: clean init info tests generate-tests report
 
 clean:
-	@echo -e "Removing $(OUT_DIR)"
-	@rm -rf $(OUT_DIR)
-	@echo -e "Removing $(TESTS_DIR)/generated/"
-	@rm -rf $(TESTS_DIR)/generated/
+	rm -rf $(OUT_DIR)
+	rm -rf $(TESTS_DIR)/generated/
 
 init:
 ifneq (,$(wildcard $(OUT_DIR)/*))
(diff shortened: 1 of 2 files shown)


piotr-binkowski:
It looks like one commit is missing a sign-off, I will merge this after its fixed

hzeller:
Ok, since the missing signed-off was on the first of two commits that can't be amended anymore after the fact, I created pull request #147

PR #152, opened 2019-09-25
tests: place signing specifier after type in integer tests

### LINKED ISSUE #151: signing in data_type
Looks like most of `/tests/generated/integers/6.11--integer_{signed|unsigned}_{type}.sv` tests have `signing` before `type`.

According to SV 2017 grammar: `signing` should be done after `type`.

```
data_type ::=
    integer_vector_type [ signing ] { packed_dimension }
  | integer_atom_type [ signing ]
  ...
```

piotr-binkowski:
Closes #151

### DIFF
--- a/conf/generators/templates/integers.json
+++ b/conf/generators/templates/integers.json
@@ -4,12 +4,12 @@
 	"template": [
 		"/*",
 		":name: integer_{2}",
-		":description: {0}{1} net test",
+		":description: {1} {0}net test",
 		":should_fail: 0",
 		":tags: 6.11",
 		"*/",
 		"module top();",
-		"{0}{1} v;",
+		"{1} {0}v;",
 		"endmodule"
 	],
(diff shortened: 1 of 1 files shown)


conf/generators/templates/integers.json
```diff
@@ -4,12 +4,12 @@
...
 		"/*",
 		":name: integer_{2}",
-		":description: {0}{1} net test",
+		":description: {1} {0}net test",
 		":should_fail: 0",
 		":tags: 6.11",
 		"*/",
 		"module top();",
-		"{0}{1} v;",
+		"{1} {0}v;",
```
tgorochowik: Are the spaces correct?

piotr-binkowski: Yes, `{0}` is specified in config as `signed `, `unsigned ` or empty string which takes care of the missing space when needed

PR #161, opened 2019-09-25
ci: remove gitignore for tests to deploy generated tests

piotr-binkowski:


.travis.yml
```diff
@@ -58,6 +58,7 @@ jobs:
         - "cp out/report.js out/deploy/report.js"
         - "cp -rf out/logs out/deploy/logs"
         - "cp -rf out/tests out/deploy/tests"
+        - "rm out/deploy/tests/.gitignore"
```
tgorochowik: This is confusing.

Can we just remove the `.gitignore` in the `tests` directory and add `tests/generated` to the main `.gitignore` in the root dir?

PR #166, opened 2019-09-26
report: add syntax highlighting

### LINKED ISSUE #162: github.io report hosting: clicking on *.sv file in report downloads them vs. loading them in frame
The report hosted on https://symbiflow.github.io/sv-tests/ considers the *.sv files as binary and sets the mime-type to `application/octet-stream`.

```
curl -I https://symbiflow.github.io/sv-tests/tests/generic/number/number_test_68.sv
HTTP/2 200 
server: GitHub.com
content-type: application/octet-stream
...
```

Thus, clicking on it will make the browser download the file instead of showing it in the frame.

It might be hard to tell the github.io what mime-type we'd like the files to be served as, as we don't have access to that webserver configuration.

We might be able to force this by adding ".txt" to the filename ? So serving `*.sv.txt`.

Or, we go the full way and generate `*.sv.html` files which we then can also pretty-print to have line-numbers (and SystemVerilog keywords highlighted?).

piotr-binkowski:
This fixes #162 and adds syntax highlighting using [pygments](http://pygments.org/).
Syntax highlighted sources look like this:
Screenshot_20190926_085418

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ generate-tests:
 report: init info tests
 	./tools/sv-report
 	cp ./conf/report/report.* ./out/.
-	cp -rfT ./tests ./out/tests/
+	find $(TESTS_DIR) -name *.sv -exec ./tools/formatter $(OUT_DIR) {} +
 
 $(foreach g, $(GENERATORS), $(eval $(call generator_gen,$(g))))
 $(foreach r, $(RUNNERS),$(foreach t, $(TESTS),$(eval $(call runner_gen,$(r),$(t)))))
--- a/conf/requirements.txt
+++ b/conf/requirements.txt
@@ -2,3 +2,4 @@ conan
 flake8
 jinja2
 tree_sitter
+pygments
--- a/tools/formatter
+++ b/tools/formatter
@@ -1,0 +1,26 @@
+#!/usr/bin/env python3
+
+import os
+import argparse
+from pygments import lexers, highlight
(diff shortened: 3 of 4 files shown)


mithro:
LGTM.

Makefile
```diff
@@ -75,6 +75,7 @@ report: init info tests
 	./tools/sv-report
 	cp ./conf/report/report.* ./out/.
 	cp -rfT ./tests ./out/tests/
+	find ./out/tests -name *.sv -exec ./tools/formatter {} +
```
tgorochowik: Do we really need to copy the test first? Do we use the non-html test in any way in the report?

tgorochowik: BTW. you shouldn't use `./tests` directly, use the `TESTS_DIR` variable.

piotr-binkowski: There is no real need for the raw `.sv` files, I will modify this to take tests from `TESTS_DIR` and only save the `.html` versions inside the `./out` directory

piotr-binkowski: Current version works without copying the files first and uses the proper variables

hzeller:
Very neat.

Can we modify the output a little with the configuration of pygments ?

  * on the top, there is some whitespace. The generated code seems to create some empty`<h2></h2>` for instance
  * If all line-numbers could contain an `<a name="L42">42</a>`, so that we can directly link from a logfile `path/to/test/file.sv:42`

PR #173, opened 2019-09-26
Add report summary

### LINKED ISSUE #164: report: Have 'total' count row of successful tests vs. overall tests.
It would be good if we'd have a row that shows the 'totals' of all tests for a particular tool, e.g. passing `117/573`. This is good for the `TL;DR` situation :)

This could be easily achieved by counting every test also towards a 'total' tag (though the row with the total count then needs to be printed a bit differently as we don't want the cells in that row to be clickable).

tgorochowik:
This PR adds a footer to the report with the total number of tests passed and the total number of tags for which all of the tests passed.

Closes #164

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -108,6 +108,21 @@
           {% endfor %}
         </tr>
       {% endfor %}
+      <tfoot>
+        <tr>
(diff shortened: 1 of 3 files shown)


hzeller:
Cool!

PR #183, opened 2019-09-27
Use anaconda packages

### LINKED ISSUE #54: Don't build tools on Travis
Currently travis is building all the tools as part of the CI. This means CI takes ~25 minutes rather than the ~5min for the run of the tests.

It would be good if the tools being used where prebuilt to make CI much faster.

piotr-binkowski:
This PR aims to replace current mix of locally built tools and system packages used in CI with packages downloaded from SymbiFlow conda [repository](https://anaconda.org/SymbiFlow/repo)
To achieve that, all tools we test should be available from that repo
Current tool status:
- [x] Verilator
- [x] Yosys - might require different package created from upstream sources
- [x] Icarus Verilog
- [x] Odin - SymbiFlow/conda-packages#40
- [x] Slang - SymbiFlow/conda-packages#43 SymbiFlow/conda-packages#48
- [x] sv2v(zachjs) - SymbiFlow/conda-packages#39
- [x] tree-sitter-verilog - SymbiFlow/conda-packages#41

Closes #54

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,6 @@ python:
 
 cache:
   pip: true
-  ccache: true
-  directories:
-    - $HOME/.ccache
-    - $HOME/.stack
-    - $HOME/.conan
 
 env:
   global:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -27,31 +22,20 @@ jobs:
...
-        - mkdir -p ~/.local/bin
-        - export PATH=$HOME/.local/bin:$PATH
-        - travis_retry curl -L https://get.haskellstack.org/stable/linux-x86_64.tar.gz | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
       script:
-        - "make slang"
-        - "make odin"
-        - "make zachjs-sv2v"
-        - "make tree-sitter-verilog"
         - "make generate-tests"
-        - "make report USE_ALL_RUNNERS=1"
```
tgorochowik: Keep this. If something is wrong with one of the packages this will detect it.

piotr-binkowski: I intend to keep this but for the moment I wanted to check how it works only with tools currently available from anaconda and with this it would fail at startup

tgorochowik: :+1:

hzeller:
The PR description should point out, that the pre-built binaries are only being used in travis to speed up CI.

(for the regular user, things should always just be built by the user unless they really want to use conda. Downloading pre-built binaries is always dangerous, in particular if it requires installing and trusting some non-standard tools such as Conda).

Personally, I'd prefer if we set up a standard way to do this: create a debian repository somewhere which can be used with standard operating system tools, can be trusted with regular key-management and it is easy to convey to people.
Given that Conda is already used currently in Symbiflow, it might be the slightly better choice at this point though.

Issue #122, opened 2019-09-20
Test quality improvement: Double-check tests that fail for all tools


hzeller:
Some test fail for all tools. This could mean they use too sophisticated SystemVerilog features. But more often than not, they contain information that is missing, such as this [number_test_23.sv](https://github.com/SymbiFlow/sv-tests/blob/master/tests/generic/number/number_test_23.sv) that is using macros without them being defined (that particular test came out of some other test-suite that just did syntax checking but all the tools are confused then because they can't resolve symbols; so here, we should define the macros).

Other tests contained other issues such discovered by @alainmarcel  and fixed in pull request #119 #116 #114 #113 #112 #111 (thanks @alainmarcel!).

So while still cranking out new tests, we should have a second look at the existing tests that fail.
I added some CSV output in the WIP pull request in https://github.com/SymbiFlow/sv-tests/commit/ce82b7964a5984dc57aea84d06e21c413706b3e8 ; with that, it is easy to look at these.

First, here is a fun overview of all the tests that are running successfully out of 496 (formulated as a SQL query for readability (and that is what I use with my personal tool for that), but `awk` or similar would work as well.)

```SQL
from csv:out/results.csv "select sum(Icarus), sum(Slang), sum(Sv2v_zachjs), sum(Verilator), sum(Yosys)"
+-------------+------------+------------------+----------------+------------+
| sum(Icarus) | sum(Slang) | sum(Sv2v_zachjs) | sum(Verilator) | sum(Yosys) |
+-------------+------------+------------------+----------------+------------+
|         122 |        198 |              202 |            183 |        117 |
+-------------+------------+------------------+----------------+------------+

from csv:out/results.csv "select 100.0*sum(Icarus)/count(*) as Ikaraus_percent, 100.0*sum(Slang)/count(*) as Slang_percent, 100.0*sum(Sv2v_zachjs)/count(*) as sv2v_percent, 100.0*sum(Verilator)/count(*) as verilator_percent, 100.0*sum(Yosys)/count(*) as yosys_percent"
+-----------------+---------------+--------------+-------------------+---------------+
| Ikaraus_percent | Slang_percent | sv2v_percent | verilator_percent | yosys_percent |
+-----------------+---------------+--------------+-------------------+---------------+
|         24.5968 |       39.9194 |      40.7258 |           36.8952 |       23.5887 |
+-----------------+---------------+--------------+-------------------+---------------+
```
[...]

PR #199, opened 2019-10-01
tests: remove invalid generic tests

piotr-binkowski:
As pointed out in #122, some generic syntax tests fail on all runners

This removes all of those failing tests


mithro:
It is unclear to me if these are actually invalid?

drom:
A big portion of these failures was because of missing `defines`.
Some of the tests (class) were probably OK:

* [ ] tests/generic/class/class_test_10.sv
* [ ] tests/generic/class/class_test_15.sv
* [ ] tests/generic/class/class_test_50.sv
* [ ] tests/generic/iface/iface_class_test_6.sv
* [ ] tests/generic/member/class_member_test_22.sv
* [ ] tests/generic/member/class_member_test_33.sv
* [ ] tests/generic/typedef/typedef_test_15.sv
* [ ] tests/generic/typedef/typedef_test_17.sv

piotr-binkowski:
Just as @drom said, most of those tests were missing `defines` like `DIGITS` or similar.

After taking a closer look at testcases listed, some of them look OK
```
class_test_10.sv
class_test_15.sv
iface_class_test_6.sv
```
Rest of testcases seem to use types, packages or variables that are not defined anywhere and `typedef_test_15.sv` looks like it is missing data type in its member declaration.

mithro:
Fix the missing defines by adding them?

piotr-binkowski:
In most cases that will fix the issues but I will need to take a closer look at cases like ```int foo = 16'b`DIGITS(bar, baz);``` I will look into adding the fixed tests back.

tgorochowik:
We agreed to just bulk remove all the failing tests from the generic set. We could be more picky about that, but most of the features there, that all the tools should actually fail on are tested in the tests for individual chapters (e.g. chapter 8 for classes) so the benefits here outweigh the losses. This way we will avoid any further reports of invalid test cases and we can start adding the fixed tests back separately (for the ones that really do make sense).

PR #219, opened 2019-10-03
travis: deploy all css and js files

tgorochowik:
#208 added a new css file for pygments, but it didn't add it to travis deploy files.

This PR makes it copy all `.js` and `.css` files from `out` to `deploy` to fix that isssue and possibly avoid similar issues in the future.


tgorochowik:
There is lots of files we don't want to copy, for example the tools we currently build in CI, some unprocessed log files (non html without the links to test files) etc.

I guess we could arrange the out directory differently to always just copy a single dir (e.g. always create the deploy directory inside it, even for local builds).

mithro:
That already create the deploy directory sounds like a good idea. Want to log a GitHub issue to do that?

tgorochowik:
Sure: #220

PR #254, opened 2019-10-07
tests: fix issue with $assert() in empty_string test

### LINKED ISSUE #241: Incorrect tests: use of $assert
The empty_string test uses an $assert function, which is not part of SystemVerilog.

piotr-binkowski:
Fixes #241

### DIFF
--- a/tests/chapter-11/11.10.3--empty_string.sv
+++ b/tests/chapter-11/11.10.3--empty_string.sv
@@ -10,7 +10,7 @@ bit [8*14:1] a;
 
 initial begin
 	a = "";
-	$assert(a == 0);
+	assert(a == 0);
 end
 
 endmodule


tests/chapter-11/11.10.3--empty_string.sv
```diff
@@ -10,7 +10,7 @@ bit [8*14:1] a;
 
 initial begin
 	a = "";
-	$assert(a == 0);
```
tgorochowik: Shouldn't it just be changed to `assert` (without the `$` character)?

piotr-binkowski: right, I thought that `assert` requires `action_block` which would make it more complicated than the `if` method. But after a closer look it turns out that it is not required and using just `assert` is the simplest solution.

PR #260, opened 2019-10-07
Limit resources the runners can take to run.

hzeller:
There are versions of `slang` that keep allocating memory until
the test-machine gets in trouble. Limit memory for sub-processes.


hzeller:
Darn, I think we should have a `make format` of sorts.

PR #270, opened 2019-10-08
Escape html in the log template

### LINKED ISSUE #268: HTML-logs not HTML escaped
The special characters for HTML are not escaped in the HTML-converted logfiles (stuff like `<, >, ", '`).

tgorochowik:
Fixes #268


tgorochowik:
Actually we have to solve that differently as we inject hyperlink tags to the logs.

PR #277, opened 2019-10-09
Modify python format checks to use yapf

### LINKED ISSUE #261: Add 'make format' for format Python code
The CI catches these, but would be good if we can have that beforehand.

piotr-binkowski:
This PR modifies `make format` command to use `yapf` to automatically adjust python code formatting
and adjusts CI to check if the code is formatted properly.
Closes #261

### DIFF
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -19,7 +19,6 @@ class BaseRunner:
     Runners must be located in tools/runners subdirectory
     to be detected and launched by the Makefile.
     """
-
     def __init__(self, name, executable=None):
         """Base runner class constructor
         Arguments:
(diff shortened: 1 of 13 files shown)


tools/sv-report
```diff
@@ -352,22 +359,21 @@ try:
...
+        report = jinja2.Template(
+            f.read(), trim_blocks=True, lstrip_blocks=True)
 
     with open(args.out, 'w') as f:
-        f.write(report.render(report=data,
-                              database=database,
-                              revision=args.revision))
+        f.write(
+            report.render(
+                report=data, database=database, revision=args.revision))
```
tgorochowik: I am not a fan of the formatting changes made by `yapf`, but if the same settings are used by symbiflow in other repos, then I guess it is ok.

tools/BaseRunner.py
```diff
@@ -45,10 +44,12 @@ def run(self, tmp_dir, params):
         """
         self.prepare_run_cb(tmp_dir, params)
 
-        proc = subprocess.Popen(self.cmd, cwd=tmp_dir,
-                                preexec_fn=set_process_limits,
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.STDOUT)
+        proc = subprocess.Popen(
```
hzeller: Can we configure it to leave 'arguments start after open parenthesis and are indended underneath' alone ? It makes for much more readable code.
The 'newline after open parenthesis' should only really be used when the arguments get too long. In any case, that should be the programmers decision.

We should make sure that formatters fix readability where needed, but don't obliterate where not needed.

mithro: Please leave the settings alone.

hzeller: My suggestions is towards readability,  wasting less vertical space and being in accordance to Google Python style-guide that I'd usually use as reference.

But if it is preferred this way in Symbiflow I won't object.

PR #279, opened 2019-10-09
report: add support for linking to testcases and chapters inside report

### LINKED ISSUE #271: Navigation: Make possible to link directly to a particular test-case
It would be good if it was possible to provide a link to a particular test-case, which is important when pointing to particular problems.

Given that things are controlled by the toplevel page, this would probably mean conveying this via a fragment, e.g.

  https://symbiflow.github.io/sv-tests/#yosys|5.6.2|keyword_uwire

(or similar, just an example. It should be human-readable though, so that it is possible for someone to hand-assemble a URL).

This then should hightlight the yosys/5.6.2 box and point the logs and test frames to the keyword_uwire tests; as if hand-navigated to that particular position.

Inversely, navigating inside the sv-test output should update the URL in the browser to the current position so that it is possible to simply copy/paste the current URL.

### LINKED ISSUE #272: Navigation: jumping laterally between tools - go to the equivalent test-case
Example: say I am currently inspecting the test-case `keyword_uwire` in tag `5.6.2` under `yosys`.

Now I want to see how that test-case looks like for the other tools.

Ideally, I'd just go sideways and click on the same row left or right of the current tool and would navigate to the corresponding test in that tool.

E.g.  If from the initial state mentioned above, I'd go to tool `verilator` while staying in row `5.6.2`, the log/src window below should also open to the `keyword_uwire` testcase directly.

This is a common use-case when comparing how different tools deal with the same situation, comparing error messages etc.

(This is probably trivially implementable once #271 is implemented as then it would just mean to replace the tool-name while leaving the rest of the fragment the same, so probably worthwhile tackling that first)

piotr-binkowski:
This PR makes it possible to link directly to the test case, links use the following format: `index.html#Sv2v_zachjs|5.7.2|number_test_71`
This also makes it possible to switch between tools and keep open the same test case

Closes #271
Closes #272

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -57,7 +57,7 @@
       <div class="logtab-bar"
            id="logtab-{{tool}}-{{tag}}-bar">
         <button class="logtab-btn logtab-close-btn"
-                onclick='toggleLog("{{tool}}", "{{tag}}", null, null)'>
+                onclick='toggleLog("{{tool}}", "{{tag}}", null)'>
           close
         </button>
         {% for test, output in tooldata["tags"][tag]["logs_sorted"] %}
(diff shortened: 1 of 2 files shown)


hzeller:
Cool, LGTM!

PR #288, opened 2019-10-10
ci: fix python code formatting

piotr-binkowski:
It appears that some files were not included in formatting fixes that were part of #277
This PR aims to fix that


tgorochowik:
Looks like a race condition with CI for various PRs.

@piotr-binkowski could you please cancel the CI for PRs that are sure to fail without this, and then re-run them once this one passes and is merged?

PR #301, opened 2019-10-11
tests: support incdirs in non-generated tests

### LINKED ISSUE #300: incdirs of preproc_test_2.sv
preproc_test_2.sv includes preproc_test_2.svh, but there is no `incdirs`.

I think `incdirs` is required bacause compiler’s current working directory is a temporary directory and the include file path as relative path from the temporary directory can't be found.

Is there any way to set valid `incdirs`?

piotr-binkowski:
This PR adds support for includes in tests that were not generated.
With this it should be possible to specify incdirs as relative paths from repository root ie. `:incdirs: tests/generic/preproc`.
If `:incdirs:` is not specified, runners will now by default set it to directory containing the test.

Fixes #300

### DIFF
--- a/tools/runner
+++ b/tools/runner
@@ -97,7 +97,7 @@ try:
         else:
             # set default values for optional metadata entries
             test_params.setdefault('files', test)
-            test_params.setdefault('incdirs', '')
+            test_params.setdefault('incdirs', os.path.dirname(test))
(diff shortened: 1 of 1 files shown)


piotr-binkowski:
I've modified the code so that it now uses `TESTS_DIR` to create include paths

PR #322, opened 2019-10-14
generators: add support for blacklisting

### LINKED ISSUE #296: utd-sv_operators - incorrect unary operators
Correct me if I am wrong, but UNARY (`~&`  and `~|`) operators were incorrectly used here:

https://symbiflow.github.io/sv-tests/#tree_sitter_verilog|utd-sv|utd-sv_operators (32:33)

```verilog
   shifter = shifter ~& result;
   shifter = shifter ~| result;
```

### LINKED ISSUE #319: Add blacklisting mechanism to imported third-party tests
>maybe we could add some kind of a blacklisting mechanism for the third-party tests, and just skip tests that are invalid?

_Originally posted by @tgorochowik in https://github.com/SymbiFlow/sv-tests/issues/296#issuecomment-541551393_

piotr-binkowski:
Closes #319
Closes #296

### DIFF
--- a/generators/path_generator
+++ b/generators/path_generator
@@ -38,6 +38,7 @@ for cfg in glob.glob(os.path.join(conf_dir, 'generators', 'meta-path',
         should_fail = data['should_fail']
         paths = data['paths']
         matches = data['matches']
+        blacklist = data.get('blacklist', [''])
 
     test_dir = os.path.join(tests_dir, 'generated', project)
 
@@ -48,6 +49,10 @@ for cfg in glob.glob(os.path.join(conf_dir, 'generators', 'meta-path',
         for match in matches:
             for f in glob.glob(os.path.abspath(os.path.join(third_party_dir,
                                                             *path, match))):
+
+                if blacklist.count(os.path.basename(f)) > 0:
+                    print("Skipping blacklisted file: {}".format(f))
+                    continue
 
                 fname = name + '_' + os.path.basename(os.path.splitext(f)[0])
                 test_file = os.path.join(test_dir, fname + '.sv')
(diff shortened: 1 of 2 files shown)


conf/generators/meta-path/hdlconvertor.json
```diff
@@ -5,5 +5,6 @@
 	"paths": [
 		["tests", "hdlconvertor", "tests", "sv_test", "*"]
 	],
-	"matches": ["*.sv"]
+	"matches": ["*.sv"],
+	"blacklist": [""]
```
tgorochowik: Can we make this parameter optional?

piotr-binkowski: done

piotr-binkowski:
I've modified `utd-sv` `.json` file to blacklist two tests that were failing for all the tools and were later verified as incorrect.
`operators` used Icarus specific operator combinations
`sparc_ffu_ctl_visctl` used reserved `logic` keyword as wire identifier

generators/path_generator
```diff
@@ -49,6 +50,10 @@ for cfg in glob.glob(os.path.join(conf_dir, 'generators', 'meta-path',
             for f in glob.glob(os.path.abspath(os.path.join(third_party_dir,
                                                             *path, match))):
 
+                if os.path.splitext(os.path.basename(f))[0] in blacklist:
```
tgorochowik: Can we change the check to check the full string? `in` will also match substrings, so if there are tests called e.g. `number`, `number-signed`, etc, it will not be possible to blacklist `number-signed` without blacklisting `number`

tgorochowik: BTW. maybe we could also use the blacklisting with the file extension to avoid similar situations?

piotr-binkowski: now it checks the full filename with extension and issue with substrings should be fixed

PR #331, opened 2019-10-16
report: don't omit tags missing from lrm.conf

piotr-binkowski:
Report generator seems to again require tags to be defined in `lrm.conf` before including them in the report file. This PR aims to fix that and to allow to use tags not defined in the configuration file.


tgorochowik:
LGTM, but needs a rebase

PR #335, opened 2019-10-17
Fix order of "Total time elapsed"

dalance:
The column of "Total time elapsed" has the wrong order than other columns.


dalance:
I rebased it.

PR #340, opened 2019-10-18
Add moore compiler

### LINKED ISSUE #339: Add moore to test suite
I absolutely love your compliance test suite. It's an absolutely excellent and necessary idea! Is there any chance to add the [moore] compiler to it? Any guidance on how to contribute would be awesome!

[moore]: http://llhd.io/

fabianschuiki:
Add the [moore] compiler to the suite of tools. Fixes #339.

[moore]: https://github.com/fabianschuiki/moore

### DIFF
--- a/tools/runners/moore.py
+++ b/tools/runners/moore.py
@@ -1,0 +1,51 @@
+import re
+from BaseRunner import BaseRunner
+
+
+class moore(BaseRunner):
+    def __init__(self):
+        super().__init__(
+            "moore",
+            executable="moore",
+            supported_features={'preprocessing', 'parsing'})
+
+        self.url = "http://llhd.io/"
+
+    def prepare_run_cb(self, tmp_dir, params):
+        self.cmd = [self.executable]
+
+        for incdir in params['incdirs']:
+            self.cmd.append('-I')
+            self.cmd.append(incdir)
+
+        tl = self.get_top_module_or_guess(params)
+        if tl:
+            self.cmd.append('-e')
+            self.cmd.append(tl)
+
+        self.cmd += params['files']
+
+    def run_subprocess(self, tmp_dir, params):
+        # Immediately fail some tests which otherwise completely break the
+        # compiler.
+        # TODO: Remove once #378 lands.
+        blacklist = [
+            "std2017/p220.sv",
+            "std2017/p221.sv",
+            "std2017/p745.sv",
+            "std2017/p341.sv",
+            "std2017/p371.sv",
+            "std2017/p759.sv",
+            "std2017/p773.sv",
+        ]
+        for arg in params['files']:
+            for bl in blacklist:
+                if bl in arg:
+                    return ("Skipping blacklisted " + arg, 1)
+        return super().run_subprocess(tmp_dir, params)
+
+    def transform_log(self, log):
+        # Strip away terminal escape codes. Moore does not yet check if the
+        # stdout is a tty that supports colorization.
+        log = re.sub(r'\x1B\[[0-?]*[ -/]*[@-~]', '', log)
+        return log
--- a/README.md
+++ b/README.md
@@ -95,3 +95,4 @@ Finally the file containing the test case and metadata should be placed in `test
 * [sv2v(zachjs)](https://github.com/zachjs/sv2v)
 * [tree-sitter-verilog](https://github.com/tree-sitter/tree-sitter-verilog)
 * [sv-parser](https://github.com/dalance/sv-parser)
+* [moore](http://llhd.io/)
--- a/conf/environment.yml
+++ b/conf/environment.yml
@@ -12,3 +12,4 @@ dependencies:
   - tree-sitter-verilog
   - slang
   - sv-parser
+  - moore
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -79,6 +79,25 @@ class BaseRunner:
         Returns a tuple containing command execution log, return code,
         user time, system time and ram usage
         """
+        result = self.run_subprocess(tmp_dir, params)
+
+        usage = resource.getrusage(resource.RUSAGE_CHILDREN)
+        profiling_data = (usage.ru_utime, usage.ru_stime, usage.ru_maxrss)
+
+        return result + profiling_data
+
+    def run_subprocess(self, tmp_dir, params):
+        """ Run the test case's subprocess
+
+        This method is called by the run method and is expected to execute the
+        command prepared in `self.cmd`. Subclasses may choose to override this
+        in order to intercept the execution of the subprocess or inject custom
+        return codes.
+
+        Arguments are the same as for the run method.
+
+        Returns a tuple containing command execution log and return code.
+        """
         self.prepare_run_cb(tmp_dir, params)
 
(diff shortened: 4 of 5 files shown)


fabianschuiki:
Currently directly installs the compiler via `cargo install`. The preferred way is to create a conda package, right?

fabianschuiki:
Will switch to conda as soon as that build passes. WIP until then.

fabianschuiki:
Previous build worked. Feel free to merge once the CI run passes!

tools/runners/moore.py
```diff
@@ -0,0 +1,49 @@
...
+            "std2017/p745.sv",
+            "std2017/p341.sv",
+            "std2017/p371.sv",
+            "std2017/p759.sv",
+            "std2017/p773.sv",
+        ]
+        for arg in params['files']:
+            for bl in blacklist:
+                if bl in arg:
+                    return ("Skipping blacklisted " + arg, 1)
```
piotr-binkowski: I think that this should be moved to the test generator [config file](https://github.com/SymbiFlow/sv-tests/blob/master/conf/generators/meta-path/hdlconvertor.json) to blacklist those tests for all runners.
Just like it was done before in similar cases where other runners misbehaved on some imported tests.

tools/runners/moore.py
```diff
@@ -0,0 +1,49 @@
...
+            self.cmd.append('-e')
+            self.cmd.append(params['top_module'])
+        else:
+            regex = re.compile(r'module\s+(\w+)\s*[#(;]')
+            for fn in params['files']:
+                with open(fn) as f:
+                    m = regex.search(f.read())
+                    if m:
+                        self.cmd.append('-e')
+                        self.cmd.append(m.group(1))
```
piotr-binkowski: Is it necessary to specify list of modules to `moore`?
If it is then maybe this part should be moved to the main `BaseRunner` class as it looks like a fairly generic feature that other runners could also use to find the top module when it is not specified explicitly
@tgorochowik ?

tgorochowik: What exactly is this doing? Are you looking for an actual module name in the test file and returning the very first one you find? Is this necessary?

@piotr-binkowski are there any other runners that need this or could work better if they had this information? I agree that it seems generic, so maybe extracting it makes sense if it could be useful anywhere else.

piotr-binkowski: Something like this could for example eliminate [Verilator warnings](https://symbiflow.github.io/sv-tests/#Verilator|hdlconv|hdlconvertor_p234) about multiple top level modules.
But after rethinking this, it appears that this automated approach can create new issues when first module is not the top one

tgorochowik: So maybe we should make the top module tag required for files with more than one module in them, and optional only for files with a single/no module?

tools/runners/moore.py
fabianschuiki: Sounds good, I will do that!

tools/runners/moore.py
fabianschuiki: > What exactly is this doing? Are you looking for an actual module name in the test file and returning the very first one you find? Is this necessary?

Indeed that's the idea. I have noticed that not all test provide a top-level module name, especially the "simple" unit tests lack that information. If no top-level module is specified, the regex just picks the first module in the file. Pretty hacky and unelegant, but gets the job done for most of the single-module tests.

fabianschuiki: > Is it necessary to specify list of modules to `moore`?

No it's not needed. If no module is passed to `moore` for elaboration, the compiler will just parse the source files. However most of the interesting business (type checking, codegen) happens upon elaboration, when the type and parameter bindings have been fully established. I suspect this is the case for many other tools as well…

Actually coming to think of it, most tools actually require you to mention the modules you are interested in (simulators, synthesizers), even though some simulators might fall back to the *"modules never instantiated = top modules"* thing.

Maybe we should require test cases to always provide one or more top-level modules in general? Then this code could live in the test generator space and tools can always assume to be given the relevant module names.

fabianschuiki:
I have factored out the lop-level module guessing into a `guess_top_module` function in `BaseRunner`. Also added the `transform_log` function to allow for easy output transformations in subclasses.

What do you think @piotr-binkowski and @tgorochowik?

tgorochowik:
Guessing the top module makes sense, however I think the check for the top module in params should be explicit (outside the guessing function). Right now it looks confusion, as it is not always really guessed, but taken from the actual test parameters.

BTW. why are you blacklisting tests that fail on moore compiler? Are the tests really broken?
If the tests are not broken but expose some bug in the compiler then they should not be blacklisted. It is ok if the compiler fails some tests with a crash - this is an useful information too.

fabianschuiki:
> Guessing the top module makes sense, however I think the check for the top module in params should be explicit (outside the guessing function). Right now it looks confusion, as it is not always really guessed, but taken from the actual test parameters.

I agree. Should I just move the check into my runner, or add a separate `get_top_module_or_guess` function to the `BaseRunner`?

> BTW. why are you blacklisting tests that fail on moore compiler? Are the tests really broken?
> If the tests are not broken but expose some bug in the compiler then they should not be blacklisted. It is ok if the compiler fails some tests with a crash - this is an useful information too.

This was in response to @piotr-binkowski's suggestion. The tests are not broken. Rather `moore` breaks horribly (infinite loop, starting to fill up the log file on disk), i.e. it crashes the entire test run. In my original PR I checked for those files in my runner and failed the test with an error message without calling the compiler at all. Should I restore that behaviour?

tgorochowik:
> I agree. Should I just move the check into my runner, or add a separate `get_top_module_or_guess` function to the `BaseRunner`?

Both options seem fine to me

> This was in response to @piotr-binkowski's suggestion. The tests are not broken. Rather `moore` breaks horribly (infinite loop, starting to fill up the log file on disk), i.e. it crashes the entire test run. In my original PR I checked for those files in my runner and failed the test with an error message without calling the compiler at all. Should I restore that behaviour?

I think we should make sure the logs are stored in a tmpfs directory (if they aren't already), there is a 4GB limit per process, so this should resolve the growing log files issue.
The infinite loop problem will hopefully be resolved by #378.

fabianschuiki:
Created a `get_top_module_or_guess` to remove confusion. Also added the moore-specific blacklist until #378 lands.

fabianschuiki:
Now with #378 landed, I'll rebase onto that and make sure my changes to `BaseRunner` work well with the new timeouts.

However I'm inclined to keep the moore-specific blacklist of the known-bad tests for now. What do you think?

fabianschuiki:
Rebased onto master and split out the changes to `BaseRunner` as separate commits. Ready to merge from my side!

tgorochowik:
@fabianschuiki could you please apply the changes needed to make PEP8 pass?

Basically `make format` should generate those changes for you, then please squash them with your initial commit.

fabianschuiki:
> @fabianschuiki could you please apply the changes needed to make PEP8 pass?
> 
> Basically `make format` should generate those changes for you, then please squash them with your initial commit.

Done. Didn't realize those checks included source code formatting. Pretty neat, I like it a lot!

fabianschuiki:
Cool, tests pass.

Issue #326, opened 2019-10-15
Create a how to document for "adding a new parser for testing"


mithro:

drom:
I guess we should discuss if it is all about SystemVerilog parsers?
Typical SystemVerilog frontend should include:

* Preprocessor
* Parser
* Semantic Analyzer 

I think that we would need to provide a way to add a new tool covering one or several of these steps. Tests need to contain information about at what stage it could be fed into the chain and at what stage it should pass/fail.

mithro:
I think that is a good start. I do think there are even a few more levels like;
 * Synthesis
 * Simulation

Issue #109, opened 2019-09-17
Parser tests?


drom:
Does this test suite target SystemVerilog Parsers or Full Compilers?

Some of the tests failing on `tree-sitter-verilog` because it is bare parser, not a preprocessor, not a compiler.

For example:

## Preprocessor

Multiple Number tests: `/tests/generic/number/number_test_*.sv` include preprocessor directives like: `DIGITS`, `WIDTH`

Other examples:

```
sv-tests/tests/chapter-5/5.6.4--compiler-directives-begin-keywords.sv
sv-tests/tests/chapter-5/5.6.4--compiler-directives-pragma.sv
sv-tests/tests/chapter-5/5.6.4--compiler-directives-unconnected-drive.sv
...
```

## Reserved words

All `/tests/generated/keywords/5.6.2--keyword-*.sv` tests suppose to fail on reserved keywords used is an identifier. I would not expect that it is Parser's job to fail in this case. Is it?

## Scope failures

There are multiple tests for scope related issues that is expected to fail by the compiler. 

Example:

```
sv-tests/tests/generated/encapsulation/8.18--inherited_local_from_inside.sv
sv-tests/tests/generated/encapsulation/8.18--inherited_local_from_outside.sv
sv-tests/tests/generated/encapsulation/8.18--inherited_prot_from_outside.sv
sv-tests/tests/generated/encapsulation/8.18--local_from_outside.sv
sv-tests/tests/generated/encapsulation/8.18--prot_from_outside.sv
```

I don't think it is the job of a Parser to fail here.

mithro:
@drom The target is to be able to test compliance around all of parsers, synthesis and simulation. Maybe we need some type of tag which describes what is needed to pass a test?

drom:
On preprocessor issue:
For all bare parsers, input code should pass via the Verilog preprocessor.
We can use: `verilator -P -E <OPTIONS> mytop.sv > mytop.prep.sv`

You are right. For tests, that supposed to fail (like Reserved words, scope analysis, etc.) we should 
add the tag, describing at what stage compiler should fail.

Nic30:
Maybe we should also separate preprocessors from parsers/compilers entirely.

tgorochowik:
This is now resolved with #355, we can now specify a type for the tests and runners and then mix and match to select what can be run on what. Thanks for the report!

PR #355, opened 2019-10-21
Add support for simulator runners

piotr-binkowski:
This PR adds support for specifying test types in the metadata - `simulation` and `parsing`, which can influence how runner handles that test case
At the moment Icarus and Verilator support simulation tests
Simulation results are verified with help of assert expressions placed in logs using `$display` statements ie. to make the tool check if `val == 5` place `$display(":assert:(%d == 5)", val);` somewhere in the `.sv` file
Additionally this adds two tags `simulation` and `parsing` to view tests only of that type

When no type is specified or is set as simulation and current runner doesn't support it then test is marked as a parsing test

### DIFF
--- a/tests/chapter-11/11.10.1--string_compare.sv
+++ b/tests/chapter-11/11.10.1--string_compare.sv
@@ -3,6 +3,7 @@
 :description: string comparison test
 :should_fail: 0
 :tags: 11.10.1
+:type: simulation parsing
 */
 module top();
 
@@ -12,7 +13,7 @@ bit [8*14:1] b;
 initial begin
 	a = "Test";
 	b = "Test";
-	$display(a == b);
+	$display(":assert:('%s' == '%s')", a, b);
 end
 
 endmodule
--- a/conf/report/log-template.html
+++ b/conf/report/log-template.html
@@ -13,6 +13,8 @@ should_fail: {{should_fail|e}}
 tags: {{tags|e}}
 incdirs: {{incdirs|e}}
 top_module: {{top_module|e}}
+type: {{type|e}}
+mode: {{mode|e}}
 files: {{file_urls}}
 time_elapsed: {{'%0.3f'| format(time_elapsed|float)}}s
 </pre>
--- a/conf/runners/vmain.cpp
+++ b/conf/runners/vmain.cpp
@@ -1,0 +1,19 @@
+/*
+ * Top level simulation driver for use with verilator
+ */
+
+#include <verilated.h>
+#include <Vtop.h>
+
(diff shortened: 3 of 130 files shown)


tests/chapter-11/11.10.1--string_compare.sv
```diff
@@ -3,6 +3,7 @@
 :description: string comparison test
 :should_fail: 0
 :tags: 11.10.1
+:type: simulation
```
tgorochowik: I think it is a good moment to also address #326 and #109 here.

So we should also add a type/stage for preprocessing (at the very least).

If no type is specified (either in the runner or the test) it should match everything (so a runner with empty types should run all tests, a test with empty types should run on all runners).

dalance: I think `:type: simulation parsing` is better because this testcase should be passed by parser too.

piotr-binkowski:
I've modified the code so that now tests without `type` specified will be used by any runner and runners without specified `test_types` will support all types of tests. 
I've also added `preprocessing` type and modified `iverilog` and `verilator` runners to support it

tgorochowik:
I think it looks good in general, here is some initial feedback:
1. The steps  (preprocessing/parsing/simulation) should not be a test tag, if you think that this kind of summary is useful, then maybe we could move it to the summary in the footer?
2. The tests that are not compatible with a given runner should be excluded from the summary, that is the runners should have different total numbers at the bottom of the report.

I also have some concerns about how the steps/types are checked at the moment. I think that there should be dependencies in them.

For example some tests should be marked with both preprocessing and parsing. It seems that with the current implementation, the tools that only support parsing will also try to run those tests.

We should have some kind relations in those, that is:
* The tests that are marked with both preprocessing and parsing should not run, if the tool only supports parsing.
* The tests that can be used for both processing and simulation should run even if the tool only supports parsing (as this makes sense with our current approach to testing simulation).

There may be more "relations" like that, we need to define and explain them.

piotr-binkowski:
I've modified the scripts to differentiate between essential features(parsing, preprocessing) optional ones like simulation which get selected only if the runner supports it and it is listed in the test requirements list.
ie. if the runner doesn't support simulation but the test lists it, it falls back to parsing only.

Essential feature verification checks if a proper subset of features is supported by runner ie. if runner supports only parsing and test also requires preprocessing then the test will be skipped, on the other hand if test requires parsing and preprocessing and tool only supports preprocessing then the test will be executed using the runner.

By default each test is handled as if it only requires parsing support and each runner by supports all the essential and optional features

PR #360, opened 2019-10-22
[WIP] tools: add scripts for automating tests of old tool revisions

piotr-binkowski:
This PR adds scripts that automate process of testing old tool revisions with current test suite

To use it, just run `tools/compare.sh <n>` where `n` is number of versions to check.
At the moment scripts just checkout all tools to a commit week ago, 2 weeks ago and so on and build version of the tool from that time
Results are saved as `out.csv` file which contains number of tests passed for each tool using code from each date


mithro:
This looks pretty good.

I would also like to support a different system in the future too. Can we look at creating a version of this script which looks at the versions available on conda?

PR #365, opened 2019-10-24
Fea/baserunner time measurement

kpyrkosz:
Added user time, system time and ram measurement to `test_params` in the runner script.

### DIFF
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -55,14 +55,15 @@ class BaseRunner:
 
         return (log.decode('utf-8'), proc.returncode)
 
-    def run_and_measure_time(self, tmp_dir, params):
+    def run_with_profiling(self, tmp_dir, params):
         """Wrapper around run method.
-        Adds elapsed time to the tuple returned by run
+        Adds user time, system time and ram usage to the tuple returned by run
         """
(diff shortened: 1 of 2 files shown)


tools/BaseRunner.py
```diff
@@ -1,7 +1,7 @@
 import resource
 import shutil
 import subprocess
-
+import time
```
tgorochowik: Do we need it here?

tools/BaseRunner.py
```diff
@@ -55,6 +55,16 @@ def run(self, tmp_dir, params):
...
 
+    def run_with_profiling(self, tmp_dir, params):
+        """Wrapper around run method.
+        Adds user time, system time and ram usage to the tuple returned by run
+        """
+
+        run_result = self.run(tmp_dir, params)
+        usage = resource.getrusage(resource.RUSAGE_CHILDREN)
+        profiling_data = (usage.ru_utime, usage.ru_stime, usage.ru_maxrss)
+        return run_result + profiling_data
```
tgorochowik: What are the results for the runners that run mostly in python? Do they make sense here? Maybe these tools (I guess only tree-sitter is already merged) should somehow overwrite this method too?

PR #367, opened 2019-10-25
Add a file of hdlconv to blacklist

dalance:
directive_verilogpp.v seems to be incorrect.
There is `unconnected_drive` without any argument, but LRM says below:

```
The directive `unconnected_drive takes one of two arguments—pull1 or pull0. 
```


tgorochowik:
It looks as though the test has already been fixed upstream:
https://github.com/Nic30/hdlConvertor/commit/a0efc8134a81eb6bfbd6f0e43cd93a4ad69f3b89

I think we could just pull their changes instead

dalance:
OK. Thanks.

PR #369, opened 2019-10-25
tests: basejump-stl: blacklist more troublesome files

piotr-binkowski:
It appears that somehow runners have issues with even more files from Basejump STL library
This aims to fix that by simply ignoring those files during test case generation


dalance:
I think the blacklisted files should be tested because the input causing trouble is very informative.
I'm trying to fix CI failure by https://github.com/SymbiFlow/sv-tests/pull/378.

tgorochowik:
@piotr-binkowski is this still needed now that #378 is merged?

piotr-binkowski:
With that PR in this is no longer needed, closing

piotr-binkowski:
On top of that we probably should create PR that does the opposite and removes the `blacklist` field from basejump config file because now it is no longer a problem if test causes tool to hang

PR #378, opened 2019-10-28
Add timeout

dalance:
Some tools seem to be stall at some tests.
This PR add timeout to each process.

### DIFF
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -1,11 +1,23 @@
+import psutil
 import resource
 import shutil
+import signal
 import subprocess
 
 
 def set_process_limits():
     """Make sure processes behave. Limit memory to 4GiB"""
     resource.setrlimit(resource.RLIMIT_DATA, (4 << 30, 4 << 30))
+
+
+def kill_child_processes(parent_pid, sig=signal.SIGKILL):
+    try:
+        parent = psutil.Process(parent_pid)
+    except psutil.NoSuchProcess:
+        return
+    children = parent.children(recursive=True)
+    for process in children:
+        process.send_signal(sig)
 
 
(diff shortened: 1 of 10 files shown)


dalance:
The longest test seems to be ariane or swerv. A minute may not be enough because both of them take about 30s. 
On the other hand, longer timeout causes to fail by Travis-CI's 50 minute timeout.
If other CI can be used, I think longer timeout is better.

piotr-binkowski:
I think that this value should be all right, I've never seen any core taking more than a minute, more like around 30s just like @dalance said. Maybe a field in metadata could be added to override the default timeout value @tgorochowik?

dalance:
I agree timeout field in metadata. Almost all tests can be processed within 10s.
So I think it is good that the default timeout is 10s and the timeout of soft-cores is a few minutes.

tools/BaseRunner.py
```diff
@@ -8,6 +10,16 @@ def set_process_limits():
     resource.setrlimit(resource.RLIMIT_DATA, (4 << 30, 4 << 30))
 
 
+def kill_child_processes(parent_pid, sig=signal.SIGTERM):
```
fabianschuiki: How about changing this to a `signal.SIGKILL` here, since `SIGTERM` can be intercepted and ignored by the process? Especially since further down you use `proc.kill()` to kill the parent.

tgorochowik: Good point

dalance: That's right.  I'll fix it.

dalance:
I added metadata `timeout`.
Now the default timeout is 10s and the timeout of soft-cores is 100s.
How about this value?

PR #397, opened 2019-11-01
Reduce visual noise in the footer

zachjs:
These changes are obviously subjective. I'm eager to receive any feedback.

Before:
image

After: (local run with fewer tools set up)
image

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -115,44 +115,44 @@
       {% endfor %}
       <tfoot>
         <tr class="report_summary_row">
-          <th class="report_summary_header" colspan="2"> Total tests passed: </th>
+          <th class="report_summary_header" colspan="2"> Total tests passed </th>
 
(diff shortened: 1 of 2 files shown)


hzeller:
Thanks, I like the lower noise!

(should these be right-aligned ? All the seconds make more sense if they are right aligned as they are numeric values. Same for megabyte.

The part/all values are really the only ones that slightly prefer centered, though I wouldn't be opposed if they end up right aligned as well if the others are)

zachjs:
@hzeller Here it is with just the footer cells left aligned. The PR has been updated with this change.
image

@tgorochowik The RAM numbers do seem pretty wild. I am running on macOS. I can take a look as to the cause.

hzeller:
@zachjs they should be _right_ aligned, not left aligned :)

zachjs:
@hzeller Whoops! Fixed.
image

hzeller:
Yay, this looks pretty cool!

PR #408, opened 2019-11-04
Dev/adding chapter 22 tests

kpyrkosz:
Wrote tests up to chapter 22.13


tests/chapter-22/22.5.1--define_and_resetall.sv
```diff
@@ -0,0 +1,18 @@
+/*
+:name: 22.5.1--define_and_resetall
+:description: The text macro facility is not affected by the compiler directive `resetall
+:should_fail: 0
+:tags: 22.5.1
+:type: parsing simulation
```
piotr-binkowski: Is this intentional?
I think that it should be the other way around, ie. `simulation parsing`, current order will cause any tool that supports both simulation and parsing to only do the parsing step

PR #409, opened 2019-11-05
Add "Average throughput passed" column

dalance:
This PR add "Average throughput passed" column.
Now the comparison of performance by "Total time elapsed" is difficult because there are many variation about pass/fail, size of test-case and type of test-case.
I think average throughout of passed test-cases may be better comparison.

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -155,6 +155,13 @@
             {{'%0.0f'| format(report[tool]["ram_usage"])}} MB </th>
           {% endfor %}
         </tr>
+	<tr class="report_summary_row">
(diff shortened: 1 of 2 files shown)


mithro:
@dalance - Adding throughput is something we had been considering and definitely makes sense. We will probably want to create some throughput based tests which are a large amount of Verilog.

hzeller:
This looks like a good idea to add, and I think it should be merged. I'd defer to people with Python experience to review the Python code for pythonicity.

dalance:
I fixed zero divide error.

tgorochowik:
Thanks!

PR #414, opened 2019-11-06
Add some tests of hdlconvertor to blacklist

dalance:
* p137.sv
`randomize_call` takes `variable_identifier_list`. So hierarchical identifier like `p.address` is forbidden.

```
randomize_call ::=
randomize { attribute_instance }
[ ( [ variable_identifier_list | null ] ) ]
[ with [ ( [ identifier_list ] ) ] constraint_block ]
```

* p552.sv
The return type of function can't take `int[$]` because `data_type` can't take `queue_dimension`.

```
function_data_type_or_implicit ::=
data_type_or_void
| implicit_data_type
```

```
data_type ::=
integer_vector_type [ signing ] { packed_dimension }
| integer_atom_type [ signing ]
| non_integer_type
| struct_union [ packed [ signing ] ] { struct_union_member { struct_union_member } }
{ packed_dimension }
| enum [ enum_base_type ] { enum_name_declaration { , enum_name_declaration } }
{ packed_dimension }
| string
| chandle
| virtual [ interface ] interface_identifier [ parameter_value_assignment ] [ . modport_identifier ]
| [ class_scope | package_scope ] type_identifier { packed_dimension }
| class_type
| event
| ps_covergroup_identifier
| type_reference
```

### DIFF
--- a/conf/generators/meta-path/hdlconvertor.json
+++ b/conf/generators/meta-path/hdlconvertor.json
@@ -7,5 +7,5 @@
 		["tests", "hdlconvertor", "tests", "verilog"]
 	],
 	"matches": ["*.sv", "*.v"],
-	"blacklist": ["p12.sv", "p940.sv"]
+	"blacklist": ["p12.sv", "p940.sv", "p137.sv", "p552.sv"]
 }


hzeller:
I wonder, should we have a comment on tests that we blacklist, so that it is easy for someon to see the reason ?

```
"blacklist": ["p12.sv", 
                  "p940.sv",
                   "p137.sv", // randomize_call takes variable_identifier_list. So hierarchical identifier like p.address is forbidden.
                   "p552.sv" // The return type of function can't take int[$] because data_type can't take queue_dimension
                  ]
```

Alternatively, is there a way to mark them as 'should fail' ?
(we should have `should_fail` markings only with a reason)

dalance:
Thanks. I think this case should be `should_fail`.
To mark `should_fail`, path_generator must be changed.
Maybe path_generator should implement `should_fail_bacause` after #449's discussion.

PR #470, opened 2019-11-23
Fix Verilator vmodel not found errors, bug438.

wsnyder:
This fixes bug 438.


wsnyder:
Note this includes an not directly related change to fix the harness crashing when there's an unknown assert: line.  This was needed to make the regressions continue running.  (Previously this issue was hidden behind the main bug this pull fixes.)

PR #471, opened 2019-11-23
Split travis into one VM per runner

mithro:
image


hzeller:
With #521 now submitted, the 'work left to do' test should now succeed.

PR #485, opened 2019-12-01
Fix some minor test issues

wsnyder:
number_test_63: Leading underscore in spec means before anything.
typedef_test_22: Fix to properly check for fwd decls.
associative/methods/traversal: Fix to avoid type mismatches.

### DIFF
--- a/tests/chapter-7/arrays/associative/methods/traversal.sv
+++ b/tests/chapter-7/arrays/associative/methods/traversal.sv
@@ -7,14 +7,14 @@
 */
 module top ();
 
-string map[ int ];
+string map[ byte ];
 byte ix;
 int rc;
 
 initial begin
     map[ 1000 ] = "a";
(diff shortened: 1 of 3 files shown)


dalance:
I think `_23_` is a valid identifier and number_test_63 should pass.
The following case can test `_23_` is not parsed as number because `time_literal` can't take identifier.

```verilog
timeunit _23_ns;
```

hzeller:
This test is testing
```
:description: Hex literal with invalid leading underscore prefix should fail.
```
... but it is now converted into something that tests underscores in decimal constants ? Was this a mistake @wsnyder ?

wsnyder:
@hzeller I changed the test to check what I think it was intended to check, then it was failing on parsers (not simulators).  Just submitted a pull request to just remove it, as don't see the value.

PR #488, opened 2019-12-04
Fix number_test_63

dalance:
`_23_` is an invalid number but a valid identifier, so this test should be passed.
To check `_23_` as number, `time_literal` can be used because it can't take identifier.


hzeller:
This test is testing
```
:description: Hex literal with invalid leading underscore prefix should fail.
```
... but testing underscores in decimals is not what we're looking for.

This was already broken in #485

PR #498, opened 2019-12-08
Fix version from scripts, and some Verilator warnings

wsnyder:
Fix to ignore some Verilator warnings.
Fix showing tool version when tool is script.

### DIFF
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -157,7 +157,8 @@ class BaseRunner:
             cmd = self.get_version_cmd()
 
             proc = subprocess.Popen(
-                cmd,
+                " ".join(cmd),  # Required to go via shell
+                shell=True,
                 preexec_fn=set_process_limits,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.STDOUT)
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -157,8 +157,7 @@ class BaseRunner:
             cmd = self.get_version_cmd()
 
             proc = subprocess.Popen(
-                " ".join(cmd),  # Required to go via shell
-                shell=True,
+                cmd,
                 preexec_fn=set_process_limits,
                 stdout=subprocess.PIPE,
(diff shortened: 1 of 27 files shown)


tools/BaseRunner.py
```diff
@@ -157,7 +157,8 @@ def get_version(self):
             cmd = self.get_version_cmd()
 
             proc = subprocess.Popen(
-                cmd,
+                " ".join(cmd),  # Required to go via shell
+                shell=True,
```
tgorochowik: Why exactly is that needed?

wsnyder: shell=True is required so that it can run a program that is not a binary, e.g. anything in perl/python with a #! such as Verilator.  Perhaps there's another way to do this but that seems the standard practice.
Once you have shell=True the join is required or the command will be broken, the python manual also discloses to pass a single space delimited line.

tgorochowik: Using `shell=True` is usually not recommended. We deal with lots of auto-generation that include third-party repositories, so I'd rather we didn't do this (see: https://docs.python.org/2/library/subprocess.html#frequently-used-arguments)

Could we override `get_version_cmd` in `Verilator.py` instead?

I think the following should work:
```
def get_version_cmd(self):
    return ["perl", self.executable, "--version"]
```

wsnyder: That doesn't work because then "verilator" won't be resolved using path.  I tried pushing using $SHELL, which works locally for me, we'll see if that passes regressions.

wsnyder:
Note Google tests failed with:
CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://conda.anaconda.org/symbiflow/linux-64/slang-v0.1_0006_gac361c58-20191211_090908.tar.bz2>

Which is unrelated to my test.  Making an arbitrary change to retest.

tools/BaseRunner.py
tgorochowik: Right, let's not dwell on it too much then, thanks for the changes!

Issue #441, opened 2019-11-16
Use Verilator --Wpedantic flag


wsnyder:
Some tests fail because Verilator is nice to users.  For example allows "reg global" because this was hitting a lot of real code before SV 2017.   Therefore some IEEE tests need to operate differently.

Please update the verilator submodule to the latest.

Then you'll see I've followed the lead of GCC and added a --Wpedantic switch, please use this flag on all of your testing of IEEE tests, but not of real projects (which might rely on extensions above IEEE).

Not sure how to only apply it to the right tests or would give you a patch.  Thanks

mithro:
I believe this was fixed in #506?

PR #506, opened 2019-12-13
Verilator correct fails and pedantic tests

wsnyder:
1. Use pedantic flags with Verilator so it avoids some good-user behavior for strict compliance.
2. Fix harness bug which was not crediting verilator for correct fails.


mithro:
I think this fixed #441 ?

Issue #449, opened 2019-11-18
should_fail should come with a reason


hzeller:
Right now, we have a label `should_fail: 1` or with zero value if it should not fail.

Often the reason why a parsing of a file should fail is subtle, so I think we should qualify that with an explanation.

I suggest to remove the old tag 'should_fail', but instead have a `should_fail_because: <msg>` tag. This tag is _only_ set in tests that actually should fail, and it is followed by a human readable explanation why.

In a transition phase (because it is a lot to change all at once), we can keep 'should_fail' in all the test-cases, and use that iff there is no `should_fail_because`. 

We then can remove all the `should_fail: 0` tags in all tests, because that is redundant information: we typically _expect_ code to not fail parsing.

This is just a change in the reporting script initially (it is an error if there is a contradictory `should_fail: 0` tag and also a `should_fail_because: <msg>` tag)

Then over time, we update all `should_fail: 1` tags with the appropriate reasonified `should_fail_because` tag.

PR #517, opened 2019-12-14
22.4--include_two_in_one_line.sv: Remove

wsnyder:
I agree the test checks what IEEE says, however no major simulator agrees this is illegal, nor do the simulators tested in sv-tests. Also this rule is extremely difficult to implement given that it must be legal to have two includes on the same line AFTER defines have been expanded (includes can be inside defines).


MikePopoloski:
Slang correctly checks for this.

MikePopoloski:
It seems that slang alone is also correctly catching that dummy_include.sv in those tests is included via angle brackets, which should not be able to find the file in the local directory. From the spec:

> When the filename  is  enclosed  in  angle  brackets  (<filename>),  then  only  an  implementation-dependent  location  containing  files  defined  by  the  language  standard  is  searched.  Relative  pathnames are interpreted relative to that location.

wsnyder:
I have no objection to a tool reporting the multiple includes, just do not see the value in requiring this check, as it forces unnecessary requirements about when defines expand. I created pull 518 to test the define case that makes this difficult.

You're right that the tests should be changed to not use <>'s, I suggest you submit a patch to fix these.

wsnyder:
Perhaps this could get merged, or I could get feedback on changes towards getting it merged?  Thanks.

Note the failure appears false, I'll re-push to retest it.

tgorochowik:
@hzeller could you take a look at the discussion?

hzeller:
I think the most value comes from tests that verify that we don't accidentally accept broken code.
If we accept code where it is impossible to misunderstand what is 'meant', this could be considered a benign extension.

So unless we can find a scenario in which not rejecting the double-include in one line can cause a problem (please comment with scenarios), I lean towards not having this test. We should not penalize behaving correctly or 'reparing' even if the syntax is somewhat off.

hzeller:
Maybe we need a new category 'strict' or 'extension' or similar, in which we complain if a parser does not follow the rules strictly, but that would be still considered ok as an extension.
Need to discuss how these should up in reports though.

So maybe we should keep this test in here for now, but add a new `:extension:` tag. 
Then, in a second step, this can be treated separately by the report until we've figured out the details.

(In a way, that is related to #449 in which a human readable reason is given why we have a specific expectation to a test failure).

So suggestion: Leave the test, add a tag `:extension: May pass, should warn; see #517` which at this point is a no-op. But if someone manually looks at the test-header they see that this might be ok and can come back to this discussion.

Then: file a separate bug in which we discuss how we deal with extensions.

hzeller:
To move this forward: for now I am fine removing this test. Let's see how we deal with extension ideas later.

@wsnyder this needs rebasing to latest head before merging it seems (or fresh pull request with that file removed, whatever is easier)

PR #519, opened 2019-12-15
Fix typedef_test_9 etc: Declare types before use as required by IEEE.

wsnyder:
Declares types before use as required by IEEE.


tgorochowik:
LGTM, thanks

PR #521, opened 2019-12-19
Don't run tests that are disabled every time.

hzeller:
Not all tests are run, e.g. tests for simulation are not
run by runners that only can do preprocessing.

Previously, these tests didn't generate any logfiles, so were
considered by the Makefile to re-run every time as the Makefile
considers the logfile as output. As a result, make never
considered the output as 'complete'.

Fix this by creating an empty logfile in that case and properly
skip these in the report generation.


hzeller:
This should help to proceed with #471 @mithro

PR #536, opened 2019-12-29
slang runner: Set correct flag for preprocessing mode, use simple compilation unit mode

MikePopoloski:
slang supports full elaboration, no need to restrict to just parsing and preprocessing.

When running in preprocessing mode, where tests don't expect to be fully compilable, use the -E flag to only preprocess.

### DIFF
--- a/tools/runners/Slang.py
+++ b/tools/runners/Slang.py
@@ -3,7 +3,7 @@ from BaseRunner import BaseRunner
 
 class Slang(BaseRunner):
     def __init__(self):
-        super().__init__("slang", "slang-driver")
+        super().__init__("slang", "slang-driver", {"preprocessing", "parsing"})
 
         self.url = "https://github.com/MikePopoloski/slang"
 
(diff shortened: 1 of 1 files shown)


mithro:
Hi,

Thank you for sending the pull request! It's awesome to see you contributing to this project. However, before I can look at this request you need to add [DCO sign-off](https://developercertificate.org/) to commits in this repository.

We require [DCO sign-off](https://developercertificate.org/) for **all** commits to this repository (including all authors). Please see the [following URL](https://github.com/probot/dco#how-it-works) to see more information and BeeWare also have a [Beginners guide to DCOs](https://pybee.org/contributing/how/dco/).

You can use `git commit --amend --signoff` and then a `git push --force` to add the signoff. You can also use `git rebase --signoff`.

Thanks!

MikePopoloski:
Sure, no problem

MikePopoloski:
Yeah, slang does not yet simulate anything. When I originally checked a few of the tests that were being excluded they didn't seem to actually rely on simulating anything, which is why I changed this. Also there are other tools, like surelog, which are set up to run these tests even though as far as I can tell surelog can't even do basic type checking, let alone simulate anything. I don't care too much either way though.

hzeller:
So if line 6 is changed back to  contain `{"preprocessing", "parsing"}`, we're ready to go ?

@MikePopoloski We probably need to document in the README which tool capabilities means what (e.g.: only enable simulation tests for tools that can `$display`)

PR #537, opened 2020-01-02
Fix basejump_stl

dalance:
I fixed some testcases in basejump_stl.

* define in bsg_cache_non_blocking_pkg.v is used in some testcases.
* module instance name is missing: bsg_sbox_ctrl_concentrate.v, bsg_stable_sort.v
* wrong associative array assignment: bsg_cache_non_blocking_mhu.sv


mithro:
@taylor-bsg - This looks right, right?

dalance:
I agree upstream fixing is better,  and removed the blacklist.

Issue #445, opened 2019-11-16
Suggested runner usability enhancements


wsnyder:
Some suggestions to make debugging tests easier using the runner:

- When creating logs/.. add a symlink to the .v input files, so a .log is adjacent to the .v that made it.

- Please add a default for OUT_DIR/TESTS_DIR/RUNNERS_DIR/CONF_DIR to be `pwd`/{out/tests/runners/conf} to make it easier to run the subcomponents standalone.  (note CONF_DIR is needed but doesn't give error like others when not set.)

- Please have a way to see all output when running a test.  I removed --quiet from runner, but it still doesn't show what is being run (e.g. command lines and outputs). Please fix this or add --verbose or something.

- Please have a option not to remove files from /tmp, to make it easier to debug.

hzeller:
#661 probably helped a little for manual debugging as it outputs the full command line first thing in the log.

PR #595, opened 2020-01-29
tools/runner: add option to keep runner work directory

piotr-binkowski:
As suggested in #445 this adds option to keep the directories in `/tmp` by setting `RUNNER_KEEP_TMP` variable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,14 @@ else
 $(OUT_DIR)/logs/$(1)/$(2).log : RUNNER = ./tools/runner
 endif
 
+ifneq ($(RUNNER_KEEP_TMP),)
+RUNNER_PARAM := --keep-tmp
+else
+RUNNER_PARAM := --quiet
+endif
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -42,8 +42,14 @@ else
 $(OUT_DIR)/logs/$(1)/$(2).log : RUNNER = ./tools/runner
 endif
 
+ifneq ($(RUNNER_KEEP_TMP),)
+RUNNER_PARAM := --keep-tmp
+else
+RUNNER_PARAM := --quiet
+endif
```
tgorochowik: Are these two mutually exclusive?

piotr-binkowski: To know which directory belongs to which test you need info logged by runner script https://github.com/SymbiFlow/sv-tests/pull/595/files#diff-3ff66653bd0ebf86aca23ca3ef75b58bR197
So you can use both options but that wouldn't be useful

tgorochowik: That makes sense, thanks for the explanation.

PR #596, opened 2020-01-31
Fix Ibex source path

towoe:
Not sure how to sync this with #590 which is updating the Ibex revision.
Included a commit for Ibex bump.


tgorochowik:
Also don't worry about #590

PR #609, opened 2020-02-02
Tests: Fix slice-equality and assignment expectations

wsnyder:
Fix some tests.
* slice-equality: Fix wrong expectation.
* chapter-8: These tests should mostly test the feature they describe and not fail if assignment-within-expression is unsupported, for which there is already a separate test.


piotr-binkowski:
I think that 11.3.6 is the test that covers this `assignment-within-expression` but only using the basic `=` operator, maybe another test could be added that uses `++`, `+=` or something else that is more advanced

tgorochowik:
Sounds good, I opened #612 for that. Let's merge this.

PR #614, opened 2020-02-04
Add equality check for escaped identifiers

towoe:
Not exactly sure if the following is correct, and also if the test makes sense.
Therefore, it would be nice the get feedback from people who are more experienced with the spec.

Following the description of Chapter 5.6.1 the backslash or following
white space is not part of the identifier.
As an example the escaped identifier `\cpu3` is the same as the
nonescaped identifier `cpu3`.

Add an escaped identifier and use it nonescaped.
Escalate the undefined usage to an error.


towoe:
Thanks for the review @piotr-binkowski. I moved the additions to a new file with the name you suggested.

PR #616, opened 2020-02-06
generators: ibex: set top module and bump ibex submodule

piotr-binkowski:


dalance:
I fixed `src_files.yml` of upstream to resolve incdir missing.
This will be merged at next version dump.

https://github.com/lowRISC/ibex/pull/582

piotr-binkowski:
@dalance Thanks for pointing that out

PR #617, opened 2020-02-06
Add earlgrey design from OpenTitan

wtatarski:


hzeller:
There is one line that is not yet passing the Python formatter; could you run `make format` and push that change ?

hzeller:
Alain found that the sequence of files in this project is not correct ? See #634

PR #620, opened 2020-02-06
Only calculate throughput for larger tests.

hzeller:
The throughput calculation makes more sense for files that are more
than just a few lines (like many of the LRM tests), so only include
tests in the throughput calculation that are at least 1KiB.

In the output, round the calculated throughput to full kibibytes/s; the
double-digit post-decimal point resolution is more accurate than
noise will allow.

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -156,10 +156,10 @@
           {% endfor %}
         </tr>
 	<tr class="report_summary_row">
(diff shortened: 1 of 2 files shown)


hzeller:
Follow-up to pull request #409 by @dalance .
Refining the accounting a bit to not include the myriad of super-small tests that (a) don't really reflect real work-load and (b) just will result in somewhat noisy output as many are processed in 0.000s

PR #621, opened 2020-02-06
Make BaseRunner.py pass 'make format'

hzeller:

### DIFF
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -33,10 +33,10 @@ class BaseRunner:
     to be detected and launched by the Makefile.
     """
     def __init__(
-        self,
-        name,
-        executable=None,
-        supported_features={'preprocessing', 'parsing', 'simulation'}):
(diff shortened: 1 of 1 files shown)


hzeller:
Interesting, so this was the result of running `make format` on my machine., but the one in travis disagrees. So maybe there needs to be more formatting configuration somewhere ?

tgorochowik:
@hzeller the configuration is here: https://github.com/SymbiFlow/sv-tests/blob/702110d2a5b1aa5d4028c453198f2b1b043c2dc5/.style.yapf

I assume you may have some global config that overwrites some yapf defaults. I guess we could add some extra settings to our local configs if it makes things easier for you (i.e. your global settings don't influence the formatting we use here)

hzeller:
What is the kind of setting that would create an extra four indent here @tgorochowik ? Maybe we should add it to the `.style.yapf` as you suggest.

Issue #138, opened 2019-09-23
Test failures when using Icarus Verilog


martinwhitaker:
I had a quick look at the test failures when using Icarus Verilog. I spotted two recurring problems:

1. Many tests do not include a top-level module. iverilog returns an error in this case, as there is nothing to elaborate. I have just enhanced the iverilog '-i' option (ignore missing modules) to also suppress this error, so using that option will fix many failures. Of course the test code will then only be checked for syntax errors.

2. A number of tests contain an 'always' construct with no waits or delays. iverilog returns an error in this case, because it will cause the simulator to go into an infinite loop. It would be best not to code tests like that.

I did spot quite a few broken tests, e.g. number_test_63.v uses invalid syntax.

mithro:
@martinwhitaker - The test suite is only very new, so if you do fine issues please log bugs.

tgorochowik:
The `-i` switch has been added a long time ago, we need to take a look at the tests which use `always` to determine if that is still a problem.

piotr-binkowski:
I've checked the tests and there are some that have this `always` infinite loop issue, I've created a small PR #622 to fix those by adding a sensitivity list to `always` blocks.

piotr-binkowski:
#622 was merged so the issue with `always` blocks should now be fixed and I think we can close this

PR #622, opened 2020-02-07
tests: fix always block to use sensitivity lists

piotr-binkowski:
As pointed out in #138 some `always` blocks were infinite loops, this PR aims to fix those cases


tgorochowik:
Does this fix #138 fully?

piotr-binkowski:
Yes, all offending `always` blocks should now be fixed

Issue #123, opened 2019-09-20
Add a sample browser-screenshot of reports to README


hzeller:
Once we have settled the new layout of the `sv-report`, we should add a screenshot to the toplevel readme so that it is easy show what to expect.

(and possibly a link to some actual output if it is generated by some continuous integration).

mithro:
FYI - The output is published at automatically built and published at https://symbiflow.github.io/sv-tests/

PR #648, opened 2020-02-20
Add a little thumbnail on the sv-tests README to easier find the resu…

hzeller:
…lts.


hzeller:
This is how it would look like: https://github.com/SymbiFlow/sv-tests/blob/90fec4fcd8be4386783d5eaac079b2bac118f30a/README.md

tgorochowik:
Is that enough to close #123 ?

BTW. If it is about making it easier to find the URL then maybe we could add the link to the repository description? The link then appears right under the repository name and is even accessible from the repository list.

hzeller:
Yes, just closed #123 
Wasn't aware that I actually opened a bug about that before :)

Issue #323, opened 2019-10-15
Incorrect tests: forward typedefs


MikePopoloski:
Forwarding typedefs are required to resolve to a valid type declaration (see 6.18). Tests that I noticed have this problem:
enum_test_0
struct_test_0
union_test_0

You might just want to make these expect to fail; other tools incorrectly allow this.

wsnyder:
I believe these are legal tests. What IEEE means is if the type is ever used it must resolve to a type.

Having forward decls that never resolve is common in real designs for the same reason that "class Foo;" in C++ often never defines or uses Foo - it's in some header for an implementation that isn't needed by this compilation due to ifdef or other reasons.

MikePopoloski:
I can't say anything about what the spec writers *meant* when they wrote it but that interpretation is not what the spec actually says. Furthermore, other tools seem to agree. Here is VCS:
```
Error-[SV-UFTD] Undefined forward typedef
testbench.sv, 6
  The definition for the forward typedef 'C' was not found in the scope where 
  it was declared.
  Provide the definition of the forward typedef.
```
Cadence:
```ncvlog: *E,TYDMSC (testbench.sv,6|10): The class specified in the typedef was never declared.```
RivieraPro:
```ERROR VCP5263 "Unresolved forward typedef C definition." "testbench.sv" 6  13```

MikePopoloski:
Fixed a while back.

PR #675, opened 2020-02-29
Various test fixes

MikePopoloski:
Things being fixed here:
* The spec says angle bracket includes only search for files provided by the language implementation, not the user's directories or the compiler's working directory.
* The spec says forward typedefs must resolve to a type within the same scope. 
* The spec says enum value initializers that use sized constants must match the size of the enum's base type.

In addition to fixing the existing tests, I added some additional coverage to catch the invalid cases mentioned above.


MikePopoloski:
This addresses issue #323

PR #687, opened 2020-03-08
support should_fail_because tag

ahegazy:
Add should_fail_because functionality solving issue #449 

specs as per the issue :
- if should_fail and should_fail_because doesn't exist, it's automatically assume should_fail: 0
- if both exist and should_fail:0, then it exits because there is a contradictory should_fail: 0 tag and also a should_fail_because: <msg> tag

* I made it backward compatible till all the tests move to the new tag, so if should_fail_because doesn't exist it'll use should_fail as before

### DIFF
--- a/conf/report/log-template.html
+++ b/conf/report/log-template.html
@@ -10,8 +10,11 @@
 <pre class="{{status}}">
 description: {{description|e}}
 rc: {{rc|e}} (means success: {{tool_success|e}})
+{% if should_fail_because|length %}
+should_fail_because: {{should_fail_because|e}}
+{% else %}
 should_fail: {{should_fail|e}}
-fail_reason: {{should_fail_because|e}}
+{% endif %}
 tags: {{tags|e}}
 incdirs: {{incdirs|e}}
 top_module: {{top_module|e}}
--- a/conf/report/log-template.html
+++ b/conf/report/log-template.html
@@ -11,6 +11,7 @@
 description: {{description|e}}
 rc: {{rc|e}} (means success: {{tool_success|e}})
 should_fail: {{should_fail|e}}
+fail_reason: {{should_fail_because|e}}
 tags: {{tags|e}}
 incdirs: {{incdirs|e}}
 top_module: {{top_module|e}}
--- a/tools/runner
+++ b/tools/runner
@@ -78,10 +78,11 @@ if args.version:
 test = os.path.abspath(os.path.join(dirs['tests'], args.test))
 
 req_test_params = [
-    "name", "tags", "description", "should_fail", "files", "incdirs",
-    "top_module", "timeout", "type"
+    "name", "tags", "description", "files", "incdirs", "top_module", "timeout",
+    "type"
 ]
 
+optional_test_params = ["should_fail", "should_fail_because"]
 test_params = {}
 
 # look for all required params
@@ -96,7 +97,8 @@ try:
             param_name = param.group(1).lower()
             param_value = param.group(2)
 
-            if param_name not in req_test_params:
+            if (param_name not in req_test_params) and (
+                    param_name not in optional_test_params):
                 logger.warning(
                     "Unsupported test param found: {} - ignoring".format(
                         param_name))
@@ -104,8 +106,9 @@ try:
 
             test_params[param_name] = param_value
 
-            if len(set(req_test_params) - set(test_params.keys())) == 0:
-                # all parameters found
+            # check all items in the req_test_params exists in the test_params.
+            if len(set(req_test_params) - set(test_params.keys())) < 0:
+                # all required parameters found
                 break
 
         else:
@@ -116,7 +119,20 @@ try:
             test_params.setdefault('timeout', "10")
             test_params.setdefault('type', 'parsing')
 
-            if len(set(req_test_params) - set(test_params.keys())) != 0:
+            # it is an error if there is a contradictory should_fail: 0 \
+            # tag and also a should_fail_because: <msg> tag
+            if ("should_fail", "0") in test_params.keys(
+            ) and "should_fail_because" in test_params.keys():
+                logger.error(
+                    "contradictory params should_fail, should_fail_because.")
+                sys.exit(1)
+            # to keep the rest of the code backward compatiple, \
+            # if should_fails not present, add the parameter
+            elif "should_fail" not in test_params.keys():
+                test_params["should_fail"] = (
+                    "0", "1")["should_fail_because" in test_params.keys()]
+
+            if len(set(req_test_params) - set(test_params.keys())) > 0:
                 missing = list(set(req_test_params) - set(test_params.keys()))
                 logger.error(
                     "Required parameters missing ({}) in {}".format(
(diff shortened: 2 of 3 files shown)


piotr-binkowski:
Hi @ahegazy, thanks for contributing. I've checked the code and in general it is looking good but there are a few things that in my opinion should be done differently.

tools/runner
```diff
@@ -78,10 +78,11 @@ if args.version:
 test = os.path.abspath(os.path.join(dirs['tests'], args.test))
 
 req_test_params = [
-    "name", "tags", "description", "should_fail", "files", "incdirs",
-    "top_module", "timeout", "type"
+    "name", "tags", "description", "files", "incdirs", "top_module", "timeout",
+    "type"
 ]
 
+optional_test_params = ["should_fail", "should_fail_because"]
```
piotr-binkowski: I think that `should_fail` and `should_fail_because` should be part of `req_test_params` and made optional by providing default values in `else` clause, ie. `0` for `should_fail` if `should_fail_because` was not provided and an empty string for `should_fail_because`.

tools/runner
```diff
@@ -116,7 +119,20 @@ try:
             test_params.setdefault('timeout', "10")
             test_params.setdefault('type', 'parsing')
 
-            if len(set(req_test_params) - set(test_params.keys())) != 0:
+            # it is an error if there is a contradictory should_fail: 0 \
+            # tag and also a should_fail_because: <msg> tag
+            if ("should_fail", "0") in test_params.keys(
+            ) and "should_fail_because" in test_params.keys():
```
piotr-binkowski: Then this check could be moved to just after the `for/else` loop and I think that it could be modified to override `should_fail` based on `should_fail_because` value, ie. if the string is not empty and `should_fail` is `0` then set it to `1` and issue a warning but if string is empty and `should_fail` is `1` then only issue a warning that `should_fail` should be replaced with `should_fail_because`

ahegazy:
@piotr-binkowski Good, what's on your mind?

tools/sv-report
```diff
@@ -271,6 +271,7 @@ def collect_logs(runner_name):
             "type", "mode", "timeout", "user_time", "system_time", "ram_usage",
             "tool_success"
         ]
+        optional_test_tags = ["should_fail_because"]
```
piotr-binkowski: In my opinion `should_fail_because` should be part of `test_tags` because we already have optional tags like `top_module` in there and I don't see any benefit of separating only `should_fail_because` like this.

conf/report/log-template.html
```diff
@@ -11,6 +11,7 @@ <h3><a href="{{runner_url}}" target="_blank">{{runner}}</a></h3>
 description: {{description|e}}
 rc: {{rc|e}} (means success: {{tool_success|e}})
 should_fail: {{should_fail|e}}
+fail_reason: {{should_fail_because|e}}
```
piotr-binkowski: I think that we could check here is `should_fail_because` is an empty string by using something like
```
{% if variable|length %}
    variable is not empty
{% else %}
    variable is empty
{% endif %}
```
and if it's not empty then show `fail_reason` but if it's empty then display old `should_fail` information

tools/runner
ahegazy: Yeah this was my initial idea, but i thought that if i added another required tag and it wasn't there the check that breaks from the loop when all tags are found would keep iterating and finally there will be an exception of missing parameter, 
Just after your comment i noticed the default value given to other tags incase they are missing,
you just made my life easier 😁

tools/sv-report
ahegazy: Yeah it was added in the case of missing parameter but we are giving it a default value so there is no need for it

ahegazy:
@piotr-binkowski updated, recheck please :D

tools/runner
```diff
@@ -96,16 +96,17 @@ try:
             param_name = param.group(1).lower()
             param_value = param.group(2)
 
-            if param_name not in req_test_params:
+            if (param_name not in req_test_params):
```
piotr-binkowski: I don't think we need extra parentheses here

tools/runner
```diff
@@ -115,6 +116,8 @@ try:
             test_params.setdefault('top_module', '')
             test_params.setdefault('timeout', "10")
             test_params.setdefault('type', 'parsing')
+            test_params.setdefault('should_fail', "0")
+            test_params.setdefault('should_fail_because', "")
```
piotr-binkowski: We should check here if `should_fail_because` is not empty and set default `should_fail` accordingly ie. when someone provides only `should_fail_because` in a test description we will still set `should_fail` to `0` and get a warning in the check further down

tools/runner
```diff
@@ -96,16 +96,17 @@ try:
...
                         param_name))
                 continue
 
             test_params[param_name] = param_value
 
+            # check all items in the req_test_params exists in the test_params.
             if len(set(req_test_params) - set(test_params.keys())) == 0:
-                # all parameters found
+                # all required parameters found
                 break
```
piotr-binkowski: Probably we should rename the `req_test_params` to something like `supported_test_params` and modify the remaining code/comments accrodingly. It's name/comment suggests that those are all required but in fact this is just a list of all parameters we look for/support which can get confusing.

tools/runner
ahegazy: yup force of habit, I always use parentheses to make my intention clear, but if you don't use it unless necessary i can remove it no problem.

tools/runner
ahegazy: yes that's what confused me at the beginning, it said required, so when i needed an optional tag i created one for it

tools/runner
ahegazy: here i set the default value if not assigned, and after the for/else, i do the check as you said in your previous [comment](for/else) [line 132](https://github.com/SymbiFlow/sv-tests/pull/687/files#diff-3ff66653bd0ebf86aca23ca3ef75b58bR132)
so if those are not set they will be set to the default case should_fail:0 and should_fail_because:""
if one of them is set the check after the for/else release appropriate warning and set shoud_fail:1 if necessary.

piotr-binkowski: I don't think that we should issue a warning if `should_fail_because` was set in the file and a default `should_fail` was used because in the long run this is how tests will be tagged ie. set `should_fail_because` for test that are expected to fail and don't bother with manually adding `should_fail`

ahegazy: yes right, i misread your 1st comment, then reread it and deleted mine, 
I did the check at first but removed it as i thought it was redundant, anyway i updated and pushed the last modifications

ahegazy:
@piotr-binkowski great thanks ..
Right now we have 1702 test case 
1582: test case with should_fail: 0
120: test case with should_fail: 1
should i open an issue for converting all of them to should_fail_because+the reason in case should_fail: 1, and remove should_fail: 0 ? or should this be added to issue #449 ? 

**I suggest that:**

- should_fail: 0 be all removed at once using simple sed command, should i make a PR for this or you see another way for it?

- and from what i noticed in should_fail: 1 files, the reason normally is written in the description: tag, so i can write a script to copy whatever in the description to the the should_fail_because: tag

piotr-binkowski:
I think it would be best to create a separate issue for converting the test cases, removing all `should_fail: 0` with a script will be all right but the `shoul_fail: 1` ones probably should be done manually or script-converted and then checked against the SystemVerilog specification, ideally to always reference the spec.

Issue #688, opened 2020-03-10
Convert all test cases from should_fail to should_fail_because


ahegazy:
As per issue #449  and PR #687 : 

Right now we have 1702 test case
- 1582: test case with should_fail: 0
- 120: test case with should_fail: 1

so we have to 
- [x] ~remove should_fail: 0 from all test cases and generators PR #689 ~
- [ x] ~Replace should_fail: 1 with should_fail_because: + reason in all test cases and generators PR #702 ~
- [ x] ~Update the README.md#adding-new-test-cases section PR #702 ~

**I suggest that:**
- should_fail: 0 be all removed at once using simple sed command.
- should_fail: 1 files, the reason normally is written in the description: tag, write a script to copy whatever in the description to the the should_fail_because: tag.

and as discussed in the comment of #449 
> but the shoul_fail: 1 ones probably should be done manually or script-converted and then checked against the SystemVerilog specification, ideally to always reference the spec.

hzeller:
So with this, we can close #449 right ?

ahegazy:
> So with this, we can close #449 right ?

well, It was closed back then in march when the #689 and #702 were merged, I don't know why it was opened again.

hzeller:
Maybe it only closed this #688 but not #449. Anyway, just closed it. Thanks!

PR #689, opened 2020-03-10
Remove should_fail: 0 from testcases

ahegazy:
As per issue  #688 this PR removes all should_fail: 0 from the testcases in directory tests

The oneliner used to remove unwanted lines: 
`for file in $(find tests -name "*.sv" -type f); do sed -i "/:should_fail: 0/d" $file;done`


piotr-binkowski:
This looks good but we also should remove `should_fail: 0` from [generators](https://github.com/SymbiFlow/sv-tests/tree/master/generators) and [generator configs](https://github.com/SymbiFlow/sv-tests/tree/master/conf/generators) so that the generated tests will be correctly tagged too

ahegazy:
@piotr-binkowski  ok, I see that there is third_party/tools/Surelog/third_party/tests/ directory containing testcases with should_fail: 0, are these the generated one?

piotr-binkowski:
Looks almost ready, I think that we also need to do something about [encapsulation tests](https://github.com/SymbiFlow/sv-tests/blob/master/conf/generators/templates/encapsulation.json) because this template generates both `should_fail: 0` and `should_fail: 1`, maybe it can be split into two templates based on `should_fail` fixed similarly to other generators

ahegazy:
> then fixed similarly to other generators.
you mean even other non-dynamic generators which have a forced should_fail: 0 or 1 should be split too for future purposes or only encapsulation and others will be done when needed?

conf/generators/templates/encapsulation-fail.json
```diff
@@ -36,9 +36,6 @@
 		["local_from_outside", "1", "b", "b_loc", "simulation"],
 		["inherited_prot_from_outside", "1", "b", "a_prot", "simulation"],
 		["prot_from_outside", "1", "b", "b_prot", "simulation"],
-		["inherited_local_from_inside", "1", "a_loc", "b", "simulation"],
-		["local_from_inside", "0", "b_loc", "b", "simulation parsing"],
-		["inherited_prot_from_inside", "0", "a_prot", "b", "simulation parsing"],
-		["prot_from_inside", "0", "b_prot", "b", "simulation parsing"]
+		["inherited_local_from_inside", "1", "a_loc", "b", "simulation"]
 	]
```
piotr-binkowski: We can now remove the 2nd and last columns and adjust the template above to always set `should_fail: 1` and `type: simulation`

conf/generators/templates/encapsulation-success.json
```diff
@@ -0,0 +1,38 @@
...
+		"    b_obj = new;",
+		"    $display(b_obj.{2});",
+		"    b_obj.fun();",
+		"end",
+		"endmodule"
+	],
+	"values": [
+		["local_from_inside", "b_loc", "b", "simulation parsing"],
+		["inherited_prot_from_inside", "a_prot", "b", "simulation parsing"],
+		["prot_from_inside", "b_prot", "b", "simulation parsing"]
```
piotr-binkowski: And similarly last column here can be also removed and `type:` can be set in the template

conf/generators/templates/encapsulation-fail.json
ahegazy: I think it would be better if we replaced the second column with the reason for 
`should_fail_because` and add the tag `should_fail_because` to the template array

piotr-binkowski: I think that for now we can just left this as `should_fail: 1` and create a separate PR for `should_fail_because` because this PR is already pretty big

PR #730, opened 2020-04-05
Fix test requiring specific fatal errors by adding runner_verilator_flags

wsnyder:
This might be controversial.

tests/chapter-6/6.19--enum_value_inv.sv tests something that Verilator has always had a check for - width violations. But as the test suite in general has many problems resulting in this message so they have been made non-fatal in the verilator runner script, so this test gets good exit status causing it to be a fail (as it's a should_fail).

I propose to fix this by allowing tool specific overrides in the test file, in this case promoting the warning to an error, so the runner will properly see the error as a should_fail and pass the test.

I only changed the verilator runner at present, an alternative would be to have the base runner class pull off appropriate runner commands.  But for now it seemed better to limit the places this was used.


wsnyder:
Note the SystemVerilog test failed as far as I can tell due to some odin compiler crash which would be unrelated to this patch.

Issue #770, opened 2020-04-21
Verilator tests broken recently w/ vbuild/vmain not found


wsnyder:
The report (https://symbiflow.github.io/sv-tests/) hadn't been updated for a while AFAIK, but I noticed now has an update.

I was all excited to see more green boxes for Verilator given recent things like classes, but alas, there's a ton of tests that were ok around a week ago and are now broken by something.   When I run it manually on master it seems to pass.

Can someone please look into what's going on?  Thanks.

E.g. chapter-6/6.13--void.sv.log under https://symbiflow.github.io/sv-tests/

```
sh scr.sh --cc -Wno-fatal -Wno-UNOPTFLAT -Wno-BLKANDNBLK -Wpedantic -Wno-context -I/tmpfs/src/github/sv-tests/tests/chapter-6 --Mdir vbuild --prefix Vtop --exe -o vmain vmain.cpp tests/chapter-6/6.13--void.sv
scr.sh: 3: scr.sh: ./vbuild/vmain: not found

```

P.S. Is there a way to see older versions of the web reports?

mithro:
@wsnyder - You bring up a good point, we should probably be publishing a history of the reports somewhere.

mithro:
@wsnyder I think they are stored in https://github.com/SymbiFlow/sv-tests/commits/gh-pages branch but not published anywhere...

tgorochowik:
> E.g. chapter-6/6.13--void.sv.log under https://symbiflow.github.io/sv-tests/
> 
> ```
> sh scr.sh --cc -Wno-fatal -Wno-UNOPTFLAT -Wno-BLKANDNBLK -Wpedantic -Wno-context -I/tmpfs/src/github/sv-tests/tests/chapter-6 --Mdir vbuild --prefix Vtop --exe -o vmain vmain.cpp tests/chapter-6/6.13--void.sv
> scr.sh: 3: scr.sh: ./vbuild/vmain: not found
> ```

This test passes for me locally - so I've created a PR to make verilator logs more verbose in the report. This should make it easier to see why it fails.

wsnyder:
The problem seems to be "ccache" is installed on some machines (namely the machine building verilator), and not all (which is why some tests randomly fail).  Defaulting to using ccache when installed was a recent Verilator change, so explains why it's breaking now.  Can you install ccache in your containers or otherwise?

If not you can suppress this with

    export OBJCACHE=

or equivalent when running Verilator, but this will waste CPU time.

PR #772, opened 2020-04-22
Make verilator script more verbose

tgorochowik:
See: #770

### DIFF
--- a/tools/runners/Verilator.py
+++ b/tools/runners/Verilator.py
@@ -21,11 +21,10 @@ class Verilator(BaseRunner):
         build_exe = 'vmain'
 
         with open(scr, 'w') as f:
+            f.write('set -x\n')
             f.write('{0} $@ || exit $?\n'.format(self.executable))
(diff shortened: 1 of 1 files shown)


wsnyder:
Thanks for debugging. BTW see a request in #445 to generally allow verbosity.

tgorochowik:
>  Please have a way to see all output when running a test. I removed --quiet from runner, but it still doesn't show what is being run (e.g. command lines and outputs). Please fix this or add --verbose or something.

The `--quiet` flag actually only controls the verbosity of the tests running progress (of what is printed to the console, having this switch makes CI faster and allows it to store fewer logs), it has nothing to do with the output that is stored in the actual report.

It seems to me that verilator was accidentally robbed of what other runners already had because of the fact that it needs several steps to run the actual tests. Most of the tools just show their whole outputs in the log files, for verilator, the compilation step was hidden - and this is what actually fails, so in my opinion it makes sense to just restore it, regardless of the verbosity options.

PR #783, opened 2020-04-30
moore: Use --syntax flag for parsing

fabianschuiki:
Rather than emitting the entire parsed AST via `--dump-ast`, use the `--syntax` flag in the parse-only test of moore.


fabianschuiki:
Excellent, thanks!

PR #292, opened 2019-10-10
Add BlackParrot RISC-V core as a test

piotr-binkowski:
Aim of this PR is to add [BlackParrot](https://github.com/black-parrot/pre-alpha-release) as a test case

Closes #61

PR #793, opened 2020-05-06
Adding BlackParrot as submodule and removing hardcoded sources

dpetrisko:
This updates BlackParrot to be tracked via submodule, which should make it easier to update when there are changes.

Also track 4 different subsets of BlackParrot builds: bp_fe, bp_be, bp_me, bp_top.


tgorochowik:
@piotr-binkowski could you take a look at this? Is the reason why it wasn't initially included as a submodule no longer valid (#292)?

dpetrisko:
May be helpful: all of the RTL in BlackParrot is in the main repo and the BaseJump STL submodule. So one solution would be to update=none in the .gitmodule and then manually update BaseJump STL in the generator script

piotr-binkowski:
@dpetrisko sorry for the delay, please update your `generators/blackparrot` file using `make format` so that CI will try to run actual tests. If timeouts are still there then I think we should do as you suggested and manually update required submodules in the generator script.

dpetrisko:
Hi @piotr-binkowski .  I made changes to .gitmodules and the black-parrot generator, but it seems that the CI automatically clones the submodules in the repo based on ./.github/add-local-submodules.sh.  What do you think is the best solution here? We could special case the .github script?

piotr-binkowski:
From what I see the `add-local-submodules.sh` script is parsing the `git submodule` output but doesn't check the `update` value. I think that best solution would be to modify this script to also look for this parameter and skip repositories accordingly.

dpetrisko:
Hi @piotr-binkowski, I have updated the script to only update submodules if update != none.  It appears to be working. However, CI is failing with the following error:
```
INFO    | Generating out/report/index.html from log files in 'out/logs'
WARNING | Tag not present in the database: earlgrey
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/sv-test-env/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/home/travis/miniconda/envs/sv-test-env/lib/python3.7/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "./tools/sv-report", line 323, in collect_logs
    tests[t_id]["first_file"] = logToHTML(t, t_html, tests[t_id])
  File "./tools/sv-report", line 173, in logToHTML
    src_relative, f, html_pat_ln, html_sub_ln, log)
  File "./tools/sv-report", line 136, in fileRefToLink
    formatSrc(fname, os.path.join(os.path.dirname(args.out), f_rel + '.html'))
  File "./tools/sv-report", line 112, in formatSrc
    with open(ifile, 'rb') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/SymbiFlow/sv-tests/third_party/cores/black-parrot/##'
"""
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "./tools/sv-report", line 392, in <module>
    results = pool.map(collect_logs, runner_names)
  File "/home/travis/miniconda/envs/sv-test-env/lib/python3.7/multiprocessing/pool.py", line 268, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/home/travis/miniconda/envs/sv-test-env/lib/python3.7/multiprocessing/pool.py", line 657, in get
    raise self._value
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/SymbiFlow/sv-tests/third_party/cores/black-parrot/##'
Makefile:125: recipe for target 'report' failed
make: *** [report] Error 1
The command "./.github/travis/run.sh" exited with 2.
```

Any idea what could be going on?

piotr-binkowski:
This is probably caused by the fact that `flist.vcs` files use `#` for comments while `flist.verilator` used `//` and generator script must be adjusted for that.

dpetrisko:
Thanks for the tip! I think everything is cleared up now

PR #794, opened 2020-05-08
Fix 6.20.2.1--parameter_unbounded.sv and 20.9--isunknown.sv

wsnyder:
Fix test problems:
    Fix 20.9--isunknown.sv: Fix width and work with 2-state simulation.


wsnyder:
@ldalek-antmicro please consider merging this too, thanks.

The Google runs failed for unassociated reasons (running this test didn't fail according to logs.).

Thanks

PR #804, opened 2020-05-10
Queue test fixes.

wsnyder:
* tests/chapter-7/queues/persistence.sv - use methods as better supported
  (Other tests check the old syntax so no new hole)
* tests/chapter-7/queues/pop_back_assing.sv - fix incorrect value

### DIFF
--- a/tests/chapter-7/queues/persistence.sv
+++ b/tests/chapter-7/queues/persistence.sv
@@ -16,7 +16,9 @@ task automatic fun(ref int e);
 endtask
 
 initial begin
-	q = { 1, 2, 3 };
+	q.push_back(1);
+	q.push_back(2);
+	q.push_back(3);
 	$display(":assert: ((%d == 1) and (%d == 2) and (%d == 3))",
 		q[0], q[1], q[2]);
 	fun(q[1]);
@@ -25,7 +27,7 @@ end
 initial begin
 	#50
 	$display(":assert: (%d == 2)", q[1]);
-	q = {};
+	q.delete();
 	#100;
 	$display(":assert: (%d == 0)", q.size);
 end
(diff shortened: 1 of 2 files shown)


tests/chapter-7/queues/persistence.sv
```diff
@@ -16,7 +16,9 @@ task automatic fun(ref int e);
 endtask
 
 initial begin
-	q = { 1, 2, 3 };
+	q.push_back(1);
+	q.push_back(2);
+	q.push_back(3);
```
ldalek-antmicro, line 21: I guess we can replace that update by assignment with `push_back` method (insert by assignment already tested in `insert_assign.sv`). LGTM.

tests/chapter-7/queues/persistence.sv
```diff
@@ -25,7 +27,7 @@ end
 initial begin
 	#50
 	$display(":assert: (%d == 2)", q[1]);
-	q = {};
+	q.delete();
```
ldalek-antmicro, line 30: Same thing as above. LGTM.

tests/chapter-7/queues/pop_back_assing.sv
```diff
@@ -14,7 +14,7 @@ initial begin
 	r = q[$-1];
 	q = q[0:$-1]; // void'(q.pop_back()) or q.delete(q.size-1)
 	$display(":assert: (%d == 2)", q.size);
-	$display(":assert: (%d == 4)", r);
+	$display(":assert: (%d == 3)", r);
```
ldalek-antmicro, line 17: This test should give same result as `pop_back.sv` (it tests same thing but via assignment). I would rather change line 14 so it access last element instead of element before, e.g. `r = q[$]`.

wsnyder: New version pushed.

PR #806, opened 2020-05-12
Tests: Turn off some ivtests which are Icarus only.

wsnyder:
This disables some of the ivtests which can only pass on Icarus because they use Icarus-only extensions.


hzeller:
Thanks Wilson! Awaiting Travis to finish, then will merge.

PR #820, opened 2020-05-18
Add tests/chapter-18 (Constrained random value generation)

tjurtsch:


tests/chapter-18/18.10--dynamic-constraint-modification_0.sv
```diff
@@ -0,0 +1,49 @@
+/*
+:name: dynamic_constraint_modification_0
+:description: dynamic constraint modification test
+:tags: 18.10 uvm
```
tgorochowik, line 4: Is there any uvm category we could put this into? The tests merged so far use for example `uvm uvm-classes` or `uvm uvm-scoreboards` etc. Could you categorize your tests too? For example `uvm umv-constraints` or `random` or however you see fit? If you think there should be more than one category for various tests that's ok too.

If you think there is no need for a category for any of the tests (or for some of them) than it is okay to stay with bare `uvm`.

PR #821, opened 2020-05-18
Add basic simulation tests for chapter 16 (Assertions)

piotr-binkowski:
This adds simulation assertion tests that use UVM.
Depends on #837

### DIFF
--- a/tests/chapter-16/16.12--property-interface-prec-uvm.sv
+++ b/tests/chapter-16/16.12--property-interface-prec-uvm.sv
@@ -1,0 +1,108 @@
+/*
+:name: property_interface_prec_test_uvm
+:description: interface property with precondition in UVM
+:type: simulation parsing
+:tags: uvm uvm-assertions
+:timeout: 60
+*/
+
+import uvm_pkg::*;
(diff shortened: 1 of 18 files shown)


tests/chapter-16/16.12--property-interface-prec-uvm.sv
```diff
@@ -0,0 +1,107 @@
+/*
+:name: property_interface_prec_test_uvm
+:description: interface property with precondition in UVM
+:type: simulation
+:tags: 16.12 uvm
```
tgorochowik, line 5: I don't think this should be tagged with `16.12` if it uses uvm. The chapter tests should be as minimal as possible.

piotr-binkowski: UVM tests are now tagged with `uvm` and `uvm-assertions` only.

piotr-binkowski:
While fixing tags I've also modified tests that should pass in simulation to be also included in parser tests and set a custom timeout value for them as I noticed that `sv_parser` can parse them correctly but it takes around 30s for each test.

PR #824, opened 2020-05-21
Adding top-module to BlackParrot builds

dpetrisko:
Trying to fix the Verilator tests, which BlackParrot is known to support


dpetrisko:
good to go!

PR #825, opened 2020-05-21
Add support for meta-tags

tgorochowik:
Meta tags are tags that combine other tags into one.
This mechanism is useful to track prerequisites for a given core/library/whatever.

As an example I've added an `uvm-req` meta tag which can be used to track prerequisites to get uvm supported. The tags that used by that meta-tag were generated using the feature analyzer tool included in sv-tests.


tgorochowik:
After some more thorough testing I think this is ready.

Note: The total number of tests (in the summary at the bottom) is not influenced by this (as this doesn't actually introduce any new tests).

PR #837, opened 2020-05-27
tools/runner: add support for use of predefined libraries

piotr-binkowski:
This PR aims to add support for using predefined libraries like UVM
As a result this will simplify tests using supported libraries by automatically adding files and include paths based on configuration stored in `conf/runners/libs.json`.
Libraries will be only added to test cases which hava matching tags ie. if test is tagged with `uvm` then you don't have to manually specify paths to UVM


piotr-binkowski:
There is only [one test](https://github.com/SymbiFlow/sv-tests/blob/master/generators/uvm) at the moment that uses UVM, I will modify it to use this new way of including UVM files.

PR #849, opened 2020-06-02
Allow out of tree runners

tgorochowik:
Thanks to these changes we can use out-of-tree runners like this:
```
RUNNERS_DIR=/my/custom/runners/ make
```

It was always the intention to be able to do this but somewhere along the way we've lost that possibility. This PR just restores this.

### DIFF
--- a/tools/check-runners
+++ b/tools/check-runners
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import os
+import sys
 import argparse
 import logging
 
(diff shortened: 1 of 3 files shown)


tools/check-runners
```diff
@@ -17,9 +18,12 @@ runner_obj = None
 
 runners_ok = []
 
+if 'RUNNERS_DIR' in os.environ:
+    sys.path.insert(1, os.path.abspath(os.environ['RUNNERS_DIR']))
+
 for runner in args.runners:
     try:
-        module = import_module('runners.' + runner)
```
piotr-binkowski, line 22: If this needed to be modified then what about [tools/runner](https://github.com/SymbiFlow/sv-tests/blob/master/tools/runner#L46)? I see that it still uses the old way of importing runner module.

tgorochowik: Absolutely right, thanks for noticing this!

PR #851, opened 2020-06-04
Add uhdm-integration runners

tgorochowik:
Marking as WIP as we don't have a conda package for this yet. Basically almost all of the tests fail for now. I've added a very simple simulation test with a one-net assignment as we really lack such simple simulation tests (the simplest tests are not marked for simulation).

Depends on: https://github.com/SymbiFlow/conda-packages/pull/97

### DIFF
--- a/tools/runners/UhdmVerilator.py
+++ b/tools/runners/UhdmVerilator.py
@@ -1,0 +1,56 @@
+import os
+import shutil
+
+from BaseRunner import BaseRunner
+
+
+class UhdmVerilator(BaseRunner):
+    def __init__(self):
+        super().__init__("uhdm-verilator", "uhdm-verilator", {"simulation"})
+
+        self.url = "https://github.com/alainmarcel/uhdm-integration"
+
+    def prepare_run_cb(self, tmp_dir, params):
+        mode = params['mode']
+        conf = os.environ['CONF_DIR']
+        scr = os.path.join(tmp_dir, 'scr.sh')
+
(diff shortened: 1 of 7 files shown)


tgorochowik:
Added a runner for the surelog-uhdm-yosys workflow as a bonus.

Also had to add Makefile targets to be able to use those runners locally.

tgorochowik:
Kokoro seems to fail for completely unrelated reasons. Rebased and pushed again in hope it helps.

tgorochowik:
Kokoro finally succeeds. @piotr-binkowski please take a look

tools/runners/UhdmVerilator.py
```diff
@@ -0,0 +1,56 @@
...
+
+            f.write(f'{self.executable} $@ || exit $?\n')
+            f.write(f'make -C {build_dir} -f Vtop.mk\n')
+            f.write(f'./vbuild/{build_exe}\n')
+
+        # verilator executable is a script but it doesn't
+        # have shell shebang on the first line
+        self.cmd = ['sh', scr]
+
+        self.cmd += ['--uhdm-ast -cc slpp_all/surelog.uhdm']
```
piotr-binkowski, line 43: I think we should use the same verilator switches [that we use in the normal runner](https://github.com/SymbiFlow/sv-tests/blob/master/tools/runners/Verilator.py#L41).
Or at least `-Wno-fatal` because without that the runner will fail even if Verilator emits only warnings.

tgorochowik: Sure, added.

PR #852, opened 2020-06-05
generators: add operator tests that use simulation

piotr-binkowski:
This PR adds new tests that check operator assignments in simulation
This also modifies the `template_generator` script itself to support providing `values` as an array of arrays to minimize amount of repetition in template configurations and adds an index variable that can be used in templates (here it was used to generate distinct names for some tests).


generators/template_generator
```diff
@@ -49,7 +49,11 @@ for cfg in glob.glob(os.path.join(conf_dir, 'generators', 'templates',
...
+    idx = 0
+
     for val in vals:
-        test_file = os.path.join(test_dir, fname.format(*val))
+        test_file = os.path.join(test_dir, fname.format(*val, idx))
         with open(test_file, "w") as wf:
-            wf.write(templ.format(*val))
+            wf.write(templ.format(*val, idx))
+
+        idx += 1
```
tgorochowik, line 59: Use:
`for idx, value in enumerate(vals)` instead

Issue #853, opened 2020-06-05
New uvm-agent tests are not legal; need UVM first


wsnyder:
The recently added uvm-agent tests use this sort of command line:

    verilator --cc ...  tests/testbenches/uvm_agent_active.sv third_party/tests/uvm/src/uvm_pkg.sv

This is not legal, IEEE requires all classes be declared before they are used.  This happened to exposed a verilator error handling bug, this is fixed in Verilator master, and Verilator now properly will give an error message.

Anyhow the uvm_pkg must be first on the command line, that is

    verilator --cc ...  third_party/tests/uvm/src/uvm_pkg.sv tests/testbenches/uvm_agent_active.sv

However I would suggest the nicer fix would be to edit each of the uvm agent sources to do a

     `include "uvm_pkg.sv"

at the top, as this makes it much easier to run the tests outside the sv harnesses, and makes it far more intuitive looking at the sources what is going on.

Thanks.

kgugala:
Hi @wsnyder thanks for catching this. We'll bump the Verilator to use the version with fixes and fix the pkg compliation order.

hzeller:
Looking at the log output, the pkg is now the first in the command-line
  https://symbiflow.github.io/sv-tests/#Verilator|uvm-agents|uvm_agent_active
Does this look good now @wsnyder ?

W.r.t using include directly: we try to keep the features tested as localized as possible so that we don't mix in testing if a tool supports include properly (though in this particular case it would not really make a big difference as uvm_pkg.sv in itself is mostly just including other stuff...).

In the above log, there are errors reported, but these look like genuine reports not related to the package order ?

wsnyder:
I think the logfile is reasonable, the error that indicates that a package is not declared is no longer present, so seems ok.

hzeller:
Alright, closing the issue.

PR #864, opened 2020-06-08
tools/runner: add library files before test files

piotr-binkowski:
This PR modifies `tools/runner` to always add library files/incdirs before the ones defined by the test itself.

This should help with #853

### DIFF
--- a/tools/runner
+++ b/tools/runner
@@ -173,14 +173,14 @@ if test_params['mode'] is None:
 
 for key in libs.keys():
     if key in test_params['tags']:
-        test_params['files'] += [
+        test_params['files'] = [
             os.path.abspath(os.path.join(dirs['third_party'], p))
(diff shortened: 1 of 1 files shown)


tools/runner
```diff
@@ -173,14 +173,14 @@ if test_params['mode'] is None:
 
 for key in libs.keys():
     if key in test_params['tags']:
-        test_params['files'] += [
+        test_params['files'] = [
             os.path.abspath(os.path.join(dirs['third_party'], p))
             for p in libs[key]['files']
-        ]
-        test_params['incdirs'] += [
+        ] + test_params['files']
```
tgorochowik, line 179: A simpler way to extend the list from the beginning without having to type `test_params['files']` twice would be:
```
test_params['files'][0:0] = [<your_comprehension>]
```
But your solution is fine too

Issue #863, opened 2020-06-08
third_party UVM submodule?


wsnyder:
sv-tests' third_party/tests/uvm is not a submodule.

Can we make it a submodule so that I and others can clone it for standalone Verilator testing etc? 

Assuming you agree this makes sense, where should it be?
  a. Chips Alliance? Upside is it's an obvious base for other CHIPS projects to include, downside is it's not really something chips "owns"
  b. Verilator?  Currently UVM is usually distributed by the simulator vendors, so this seems appropriate, and also we will eventually want some `ifdef VERILATOR lines in there to improve some performance issues.
  c. Other?

Regardless of a/b/c I would ideally like it in a place where I have commit permission so can maintain a branch holding the subset that Verilator can currently parse (all but ~63 lines).

tgorochowik:
At the moment `third_party/tests/uvm` is just the code extracted from the archive Accellera provides (as mentioned in this [commit message](https://github.com/SymbiFlow/sv-tests/commit/b70882180c1df9f32e22c18bc677403ed272612b)).

Or do you want all the individual uvm tests to be moved out of the tree? For now they are in the `tests/` directory as they aren't really "third party", see an example [here](https://github.com/SymbiFlow/sv-tests/blob/6efc94ec4ba45c24bf4d97619310a4a1f097b383/tests/chapter-16/16.10--property-local-var-uvm-fail.sv).

I guess this could be moved to a separate repository if it would be in any way useful. One issue we would have to solve is moving the test generators out of the tree too as some of the uvm tests are generated from [templates](https://github.com/SymbiFlow/sv-tests/blob/686ba0807a28a3e614a09120b1ee55c9bcb9072d/conf/generators/templates/uvm-classes_0.sv).

wsnyder:
I'm interested in having third_party_tests/uvm as a repository (like easyUVM), so it can be a submodule of sv-tests, of verilator_ext_tests, and of people using it in production for their own code (e.g. eventually SweRV will have tests and need the submodule),

I think the main question is where should it live?   E.g. if I add a verilator/uvm would you be willing to point to it?

wsnyder:
#865 if acceptable is a pull to do what I'm suggesting.

tgorochowik:
The PR is merged, closing, thanks!

PR #865, opened 2020-06-08
Add UVM as a submodule

wsnyder:
This adds UVM as a submodule as described in #863.

I made the repo it points to with every version of UVM with appropriate dates and tags, so should be a reasonable reference.


tgorochowik:
After some internal discussion we came to a conclusion that we would prefer to have it next to sv-tests (so symbiflow _for now_)- as you know we are working on adding UVM support for verilator too so it's important that we can modify the code base too (even if only on some branches).

Would you mind moving it to https://github.com/SymbiFlow/uvm ? We invited you there as an outside collaborator with write access so you should be able to push there too.

CC @kgugala

wsnyder:
Think it's all in place - we'll hope the regressions pass ;)

You might want to turn off wiki/issues/pull under Settings on github for SymbiFlow/uvm.

PR #866, opened 2020-06-10
Fix runners logs decoding

tgorochowik:
Some runners seem to throw non utf-8 characters in their log for some
specific tests. Ignore the invalid characters to allow the test suite to
continue.


wtatarski:
LGTM!

PR #875, opened 2020-06-17
earlgrey: dynamically scan for source files

### LINKED ISSUE #764: The earlgrey test should scan for source files dynamically
See #748 and the fix in #761 

Ideally we should use a python `glob` in the test generator to find all the needed files (instead of having the list of files in a config file):
https://github.com/SymbiFlow/sv-tests/blob/d60466dd1b2b778bfd67e425b4082378f227ba21/generators/earlgrey#L42-L46

Note that we may actually need to use both approaches to keep including the uvm stuff.

Otherwise it will keep breaking with changes to the submodule (another question is why the bumped submodule was even merged by dependabot)

wtatarski:
Fixes #764


generators/earlgrey
```diff
@@ -36,14 +37,31 @@ sources = ''
...
         line = line.strip()
-        incs.append(os.path.dirname(os.path.join(earlgrey_path, line)))
-        sources += os.path.join(earlgrey_path, line) + ' '
+        if "+incdir+" in line:
+            line = line.replace('+incdir+', '')
+            incs.append(
+                os.path.dirname(os.path.join(earlgrey_conf_path, line)))
+            continue
+        elif "+define+" in line:
+            continue
```
tgorochowik, line 61: I really think we need a mechanism to use those defines if possible. These should be somehow passed to the runners (via another `:marker:`) and selected runners should be able to use it. I think it is ok if we only add it to verilator at first.

generators/earlgrey
```diff
@@ -11,6 +11,7 @@ templ = """/*
 :files: {0}
 :incdirs: {1}
 :tags: earlgrey
+:defines: {2}
```
tgorochowik, line 14: That should do!

What is the accepted format of a define?

I think that we should have at least one static/non-generated test that demonstrates this.

PR #877, opened 2020-06-17
tests: add minimal operator tests

piotr-binkowski:


tests/chapter-11/simple/11.4.11--simple_cond_op-sim.sv
```diff
@@ -0,0 +1,11 @@
+/*
+:name: simple_cond_op_sim
+:description: simple ?: operator simulation test
+:type: simulation parsing
+:tags: 11.4.11
+*/
+module top(input a, output b);
```
tgorochowik, line 7: Wouldn't it be better to use local nets instead of input/outputs so they can be easily assign some values to them and use our assertion mechanism so we can actually verify if the simulation was correct?

piotr-binkowski: I was aiming for tests that use only minimal set of features (module, `assign` statement, input/output `wire` and the operator itself). Existing operator tests use `int` variables, `initial` blocks and `$display` statements so I thought that something minimal would be useful.
Maybe those tests should be relabeled as parsing only?

tgorochowik: Ok, this makes sense. Is there a way we could verify whether these assignments actually work?

piotr-binkowski: Checking whether assignments are correct or not would require adding at least `reg` and `$display` keywords and that increases complexity considerably (compared to a minimal test case).
Maybe best solution here would be to mark those tests as `parsing` (after all results can't be verified so we only know that the tool processed that file without failing) and modify existing operator tests to use only a minimal set of features (no `int` but `wire` and `reg` and only minimal `$display` formatting).

tgorochowik: Let's keep it then. We can have some tests that just run the simulation and some tests that actually verify it.

I would not change the type of the test to parsing only as this would exclude this for being used for some tools (e.g. the upcoming surelog->uhdm->verilator flow).

Could we at least add a comment about that in the description of the tests? I mean something saying that this test is intentionally not verifying the results of the simulation bla bla bla.

piotr-binkowski: I've updated descriptions for the tests

PR #879, opened 2020-06-18
Tests cleanup

### LINKED ISSUE #862: Class tests need UVM first to be legal IEEE
Thanks for adding more SV tests, these will be helpful as tools support more of UVM.

Some cleanup is needed as many of the tests are not IEEE legal.  Taking for example class_member_test_2:

    class myclass;
     extern task subtask(arg_type arg);
    endclass

This is not legal as arg_type is not declared.  This fails on all of the closed-source simulators.

Could these be cleaned up please, and ideally have a process to check tests against some simulator (parsing is insufficient) before they hit master?  

Thanks

wtatarski:
Removed/modified tests which have illegal IEEE syntax or are already
covered in other tests.

Fixes #862


tests/generic/preproc/preproc_test_12.sv
```diff
@@ -6,4 +6,4 @@
 */
 `define LONG_MACRO(
     a, b=2, c=42) \
-a + b /c +345
+a + b /c +345;
```
tgorochowik, line 9: do we really need it in a macro?

wtatarski: Actually, I decided to remove this test completely, since the whole macro is poorly formatted.

PR #900, opened 2020-06-29
Migrate all remaining "should_fail" statements to "should_fail_because"

tjurtsch:
This is related to https://github.com/SymbiFlow/sv-tests/issues/449

This PR does following things:
- remove all "should_fail: 0" statements because they are not needed (tests are assumed non-failing by default)
- change "should_fail: 1" tests to "should_fail_because" and add human-readable explanation

### DIFF
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -10,8 +10,8 @@ templ = """/*
 :description: Test imported from ivtest
 :files: {1}
 :incdirs: {4}
-:should_fail: {2}
 :tags: ivtest
+{2}
 {3}
 {5}
 */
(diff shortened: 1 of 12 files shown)


tjurtsch:
@tgorochowik PTAL

generators/ivtest
```diff
@@ -216,19 +214,16 @@ for l in ivtest_lists:
                 continue
 
             type_ = ''
-            for t in type_should_fail:
+            for t in type_list:
                 if re.match(t, line[1]):
-                    should_fail = 1
                     type_ = ':type: simulation'
```
piotr-binkowski, line 219: I think those tests should still be marked with `should_fail_because` because `CE` and `RE` are test types that are supposed to fail:
https://github.com/steveicarus/ivtest/blob/1fa31c0e881290cddb2363b6ef52256fffc56337/regress-ivl1.list#L40-L44

tjurtsch: Ok, so what the "because" for CE and RE tests, respectively?

tjurtsch: Ok, I get it, compile error and runtime error.
I'll fix this.

tjurtsch:
@piotr-binkowski PTAL

PR #452, opened 2019-11-20
kokoro: Run everything, then display first failure.

mithro:
Run everything that can be run in parallel, then output a single failure.

Signed-off-by: Tim 'mithro' Ansell <me@mith.ro>

PR #911, opened 2020-07-07
cgroups: Provide more default memory.

hzeller:
* When run with cgroups, provide a little more memory to
   run the runners (3GiB vs ~2GiB). Can now also be set by
   CGROUP_MAX_MEMORY variable.
 * Use the USE_CGROUP variable everywhere (some places still
   used a hard-coded value)
 * Make sure the memory limit is initialized even if the
   cgroup already exists: this allows to experiment with
   different memory settings with each run.
 * don't create $(1)-cg: target twice (else/endif was missing)
 * Somewhat independent, but while at it: provide a little longer
   timeout to black-parrot.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ THIRD_PARTY_DIR ?= ./third_party
 GENERATORS_DIR ?= ./generators
 
 USE_CGROUP := ${USE_CGROUP}
+CGROUP_MAX_MEMORY ?= 3221225472  # 3GiB
 
 export OUT_DIR
 export CONF_DIR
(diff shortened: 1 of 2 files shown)


hzeller:
The build-failures look like glitches
  * Travis didn't one test because it ran into a temporary git issue
  * The Kokoro build had a single issue with a FAIL: Slang/chapter-9/9.7--process_cls_self.sv

Makefile
```diff
@@ -68,17 +69,17 @@ define runner_cg_gen
 ifneq ($(USE_CGROUP),)
```
mithro, line 69: There was a reason I wasn't using an `else` block here -- but I can't remember why now.....

hzeller:
Found the pull request. And now I remember that I commented on that - https://github.com/SymbiFlow/sv-tests/pull/452
Does that ring a bell what the problem was ?

PR #912, opened 2020-07-08
Run Kokoro tests without cgroups.

hzeller:
We're running with -j1, so only one test at at time is running
anyway.

### DIFF
--- a/.github/kokoro/run.sh
+++ b/.github/kokoro/run.sh
@@ -38,7 +38,8 @@ echo "Running tests"
 echo "----------------------------------------"
 (
 	set +x +e
-	export USE_CGROUP=sv-tests
+	#export USE_CGROUP=sv-tests
+	free -h   # Let's see how much memory we have
 	source "$HOME/miniconda/etc/profile.d/conda.sh"
 	hash -r
 	conda activate sv-test-env
--- a/.github/kokoro/run.sh
+++ b/.github/kokoro/run.sh
@@ -38,8 +38,11 @@ echo "Running tests"
 echo "----------------------------------------"
 (
 	set +x +e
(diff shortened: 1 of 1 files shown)


mithro:
@hzeller -- Actually, I remember what happens now without cgroup stuff. The OOM kills the parent job so the whole build errors and you don't get any output....

hzeller:
Ah, I see. So we should use cgroups but get the allowable size as close as possible to the physical limit.

In this run, I've added a `free` to see how much memory we've got to work with. I can't see the ongoing log, so I'll wait until that finishes and then update this change to set the memory limit accordingly.

hzeller:
The `free` call shows that we have plenty of physical space (though not clear if we can use all of this, we might be ulimit-jailed in in some form).

```
              total        used        free      shared  buff/cache   available
Mem:            14G        166M        5.5G        8.6M        9.0G         14G
Swap:          1.0G          0B        1.0G
```

All the current test passed without cgroups, so maybe we should try without for a while until we run into issues. If so, the next version could be something like
```
export CGROUP_MAX_MEMORY=$(free -b | awk '/Mem:/ { printf("%.0f\n", $2 * 0.8);}')
```

hzeller:
Looks like this improved the situation: Now Surelog shows green for [Ariane](https://symbiflow.github.io/sv-tests/#Surelog|ariane|ariane) and [Earlgray](https://symbiflow.github.io/sv-tests/#Surelog|earlgrey|earlgrey).

Peak usage (bottom of the page) of Surelog is about 2.4GiB but slang (7.5GiB) sv2v (5.5GiB) peak at higher, but stays within the 70% of available space of the CI machine.

PR #914, opened 2020-07-10
Show bargraph for partially passing tests.

hzeller:
For all test categories that comprise of a bunch of tests,
we currently show 'yellow' if they only pass partially.

Given that there are some of these where there are hundreds
or even thousands of tests lumped together, this is a very
coarse grained representation.

Changed this to a bargraph representation: now the percentage
of the box that represents passing tests is green, the rest
red.

We're loosing a happy yellow color in the output, but gaining
much easier insight of fraactions on tests on first sight.

bargraph-svtests


mithro:
That looks super cool!

Issue #915, opened 2020-07-10
Classification of test to should fail. Wrong application of json file ?


hzeller:
The utd-tests are currently mostly classified as failing.
![utd-classified-as-should-fail](https://user-images.githubusercontent.com/140937/87112719-699fa180-c221-11ea-8586-9833de98cf1c.png)

.. with a suspicious should-fail message *imported from yosys repository contains intentional syntax error*

This string shows up only in `conf/generators/meta-path/yosys-errors.json`. Is the wrong json file applied while generating the utd tests ?

Assigning to @tjurtsch as he was last editing the yosys json file (though the problem is probably not in the file but in the generator).

hzeller:
Ok, assigning to @tgorochowik for now as @tjurtsch does not show up yet in the bug-assignment list (do we need to add him @mithro ?)

tgorochowik:
One extra thing I've noticed is that we have an empty `should_fail_because:` for tests that are expected to pass. This doesn't seem to break anything but looks confusing, so we should probably get rid of that too.

hzeller:
This example above now moved to basejump:
![json-include](https://user-images.githubusercontent.com/140937/87158171-2a4e7080-c274-11ea-9001-5c936a6119e8.png)

Could this be because here all json files are processed in *.json order, which is essentially random depending on how the inodes are arraged ? Should we use *.json in the first place or maybe more specific ?

piotr-binkowski:
I just noticed that this may be caused by the fact that we only clear `should_fail_because` once here https://github.com/SymbiFlow/sv-tests/blob/master/generators/path_generator#L28 which is outside of the json processing loop so next json file processed will use the message from previous one even if it shouldn't fail @tgorochowik @tjurtsch

hzeller:
That looks like it explains what we see @piotr-binkowski

PR #917, opened 2020-07-10
conf/report/log-template: make should_fail_because conditional

tjurtsch:
We don't want empty should_fail_because in logs

Related to https://github.com/SymbiFlow/sv-tests/issues/915

Signed-off-by: Tomasz Jurtsch <tjurtsch@antmicro.com>


tjurtsch:
@tgorochowik PTAL

PR #918, opened 2020-07-10
[dv:axi_vip] Addign initial version of AXI VIP

rasmus-madsen:
Initial version of AXI VIP I need to check location is ok and that files has been scrubbed ok before converting to actual PR

Signed-off-by: Rasmus Madsen <rasmus.madsen@wdc.com>

### DIFF
--- a/sv/axi_vip/axi.f
+++ b/sv/axi_vip/axi.f
@@ -1,0 +1,28 @@
+# Compile list for axi
+#project defines
++define+ID_WIDTH=16
++define+ADDR_WIDTH=64
++define+BYTE_WIDTH=32
++define+USER_WIDTH=4
+#folder includes
++incdir+./modules/vip/axi/tests
++incdir+./common/sim/vip_lib/sys_uvc
++incdir+./common/sim/vip_lib/axi_uvc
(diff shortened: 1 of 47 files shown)


rasmus-madsen:
hi @kgugala 
could you take a quick look at this and see if it looks as expected and if the location is ok?
I will do a second check of the files over the weekend to make sure the files are clean.

hzeller:
also @tgorochowik is a good person to give feeback.

As location, I suggest something like `third_party/tests/axi_vip/`

rasmus-madsen:
> also @tgorochowik is a good person to give feeback.
> 
> As location, I suggest something like `third_party/tests/axi_vip/`

@mgielda or @kgugala  do you have any comments on this if not I will update the location and update the PR

kgugala:
the `third_party/tests/axi_vip/` seems fine. We still need to go through all the code in the PR, but moving it to `third_party` is sth we should do

rasmus-madsen:
> the `third_party/tests/axi_vip/` seems fine. We still need to go through all the code in the PR, but moving it to `third_party` is sth we should do

np closing PR and opening a new one

PR #920, opened 2020-07-10
Use bargraph representation also in summary line.

hzeller:
... with a slightly nasty calculation directly in the template.

summary-bargraph


hzeller:
now the bargraphs at the bottom look very smushed together. I think a border will help with that. Updated PR.
summary-bargraph2

Issue #913, opened 2020-07-09
make generate-tests fails for earlgrey out of the box


hzeller:
On a default install with all the dependencies installed [mentioned in the README](https://github.com/SymbiFlow/sv-tests#running), `make generate-tests` will fail, specifically:

```
$ make generate-earlgrey 
./generators/earlgrey earlgrey
Traceback (most recent call last):
  File "./generators/earlgrey", line 42, in <module>
    subprocess.call(
  File "/usr/lib/python3.8/subprocess.py", line 340, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1702, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'fusesoc'
make: *** [Makefile:138: generate-earlgrey] Error 1
```

Reason is, that there are a few requirements that need to be installed; after doing that with
```
pip3 install --user -r ./third_party/cores/opentitan/python-requirements.txt
```

The above `make generate-earlgrey` call works without errors.

Getting these requirements should be part of the `./generators/earlgrey` script. Is it possible for the script to install these dependencies locally in a scratch area without messing with the users' `~/.local` ?
I leave it up to the Python experts how to solve this best.

tgorochowik:
Using a scratch area is a good idea (I guess we will have to use virtualenv for that), but I think that expanding the README is a good first step - the fusesoc dependency is included in our requirements.txt too (and that's what the CI currently uses).

PR #922, opened 2020-07-10
Add missing `-r` to pip install call.

hzeller:
Issue #913


hzeller:
The README is still targeting the regular local building, which is what someone using the tests is more interested in.

The conda stuff we mostly use for CI to get all the things; but it might be also useful to document, I agree (but not this change).

mithro:
SGTM!

PR #929, opened 2020-07-11
Normalize CSV output and emit relevant data for each (test, tool) tuple.

hzeller:
Previous CSV file had one column per tool and in each of these only the
outcome of the test: True or False.

Now, have one row per (test, tool) tuple and output more relevant
information.

  * `TestName`, `Tool`  *the selector*
  * `Pass`  *{True or False}*
  * `Tags`  *associated tags with the TestName; space separated*
  * `InputBytes` *Bytes the input (System)Verilog files add up to*
  * `ExitCode` *Exit code of the tool. Special value 71 for timeout (`71meout` :))*
  * `AllowedTimeout`  *Timeout allowed for this test*
  * `TimeUser`, `TimeSystem`, `TimeWall`
  * `RamUsageMiB`

We are not emitting all associated filenames anymore per test - this was partially
a huge list for some of the tests, and not terribly useful. The TestName is unique.

Also: file content of report.csv changed, so someone who was relying on the old format
is breaking. However I'd say someone who is well-versed enough to use the CSV, will
quickly fix things, so we should not have an ugly backward compatible version.

### DIFF
--- a/tools/BaseRunner.py
+++ b/tools/BaseRunner.py
@@ -128,7 +128,7 @@ class BaseRunner:
             proc.kill()
             proc.communicate()
             log = ("Timeout: > " + str(timeout) + "s").encode('utf-8')
-            returncode = 1
+            returncode = 71  # 71meout :) - something easy to grep for
 
         invocation_log = " ".join(self.cmd) + "\n"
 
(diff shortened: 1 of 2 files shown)


hzeller:
This will now help to easily find the tools that timed out on CI, so that we can adapt the timeout values and avoid glitching test results (the new timeout exit code 71 (for `71meout`) makes it easy to find.

Also it is easy to find minimum/maximum time or memory usage as well as comparing tools, e.g. how they compare in processing time vs. input file size:
size-time

mithro:
Ohhh -- very pretty.... Would it make sense to automatically generate and publish something like that as part of sv-tests?

hzeller:
I think once we have some metrics that playing with the data shows make sense to look at, we should do that!

Issue #931, opened 2020-07-12
Report page: move logfile buttons to the bottom of the page ?


hzeller:
The current index.html is about 10MB, so it will take a while to load.

The visible part (the table) is actually emitted at the end of the file while at the beginning all the hidden logfile button links are. Can these be moved to the end of the page and maybe improve incremental loading and displaying of the result page ?

tgorochowik:
Fixed in #933

PR #933, opened 2020-07-12
Report-Page: Add a headline with link back to sv-tests.

### LINKED ISSUE #930: Report page should have a link back to sv-tests project
Just looking at the report page output https://symbiflow.github.io/sv-tests/ leaves the user wondering where this is coming from.

So we should have a headline linking back to this repository.

hzeller:
Also: move logfile divs to the bottom of the page to hopefully
improve incremental loading experience.

Fixes #930 #931

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -42,47 +42,9 @@
     }
   </style>
 
-  <div id="logfile-outer">
-    <div class="iframe-wrap">
-      <iframe class="iframe-log" name="log-frame"></iframe>
-    </div>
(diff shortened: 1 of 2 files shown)


hzeller:
Q: for @tgorochowik : is it ok to move the logfile divs to the bottom or could there be some something-needs-to-be-defined-first-before-referenced issue ? I suspect not, just double-checking.

tgorochowik:
@hzeller the order of the divs should not matter (I'm on vacation currently so sorry if I'm slow to respond)

PR #936, opened 2020-07-13
Fix Ariane test: there were a few missing files.

hzeller:
Only Verilator sptted it, as it is capable of doing Ariane,
but is also the only one that resolves instantiations.

The diff is not very helpful as it is one big line. The files added were
  * fpga/src/apb_timer/apb_timer.sv
  * fpga/src/apb_timer/timer.sv
  * tb/common/mock_uart.sv


hzeller:
This is the verilator result that currently fails and will be fixed with this https://symbiflow.github.io/sv-tests/#Verilator|ariane|ariane

hzeller:
Looks like Kokoro is stuck for the last 14h. But Travis didn't complain at least.

tgorochowik:
We should merge this, but we should also find a way to update the files list automatically. Could we use fusesoc for Ariane the same way we do for earlgray?

PR #938, opened 2020-07-15
Reduce the amount of data generated when collecting logs.

hzeller:
The multiprocessing seems to run into trouble on the
CI machines when transferring large amounts of data.

This is at the expense of making the html template more ugly though...

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -62,13 +62,13 @@
           {% for tool, tooldata in report|dictsort %}
           <th class="report_table_result {{ tooldata["tags"][tag]["status"] }}
             {% if "test-na" not in tooldata["tags"][tag]["status"] %} test-cell {% endif %}"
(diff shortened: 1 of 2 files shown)


hzeller:
Looks like reducing the data transmitted allows for a smooth CI execution.

hzeller:
Reason for the large data problem: the parallel parts run in separate processes through the logs, and also happend to collect all logfile contents in a map. When joining the processing pool, Python was transferring the data back from the forked children via some sort of IPC and crashed when the data was too large (at least this is what the stacktrace hinted at).

It was working locally, but the Kokoro CI was not happy, and in fact never finished as it was leaving some process behind waiting patiently for data to be shuffled back...).

The reduced amount of data by not including the logfile content (in the second commit in particular, the first only reduces by a small string per test) made the Kokoro CI succeed.

PR #940, opened 2020-07-16
Recursive function test

M1Fream:
A test to validate recursive automatic functions. Contains a recursive implementation of factorial and 5 test cases.


tgorochowik:
Travis fails to report the status on this PR for some reason. Going into the details of travis shows that all the individual jobs actually pass with no problems. Let's merge it.

PR #953, opened 2020-07-22
README: Normalize shell snippets

tgorochowik:
Use `$` consistently before all shell commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ The report generated from the last passing master build can be viewed [on a dedi
 
 Initialize the submodules:
 
-```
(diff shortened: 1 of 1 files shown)


hzeller:
mmh, I tend to actually like it to be the other way around. That way, it is easy to copy/paste. With `$` in front, it forces the user to manually copy each line, leaving out the start-$

tgorochowik:
@hzeller I don't mind either. Adding `$` makes sense when there are snippets for various interpreters which is not the case here. So let's just make them consistent - I removed `$` everywhere and pushed the new changes.

Issue #932, opened 2020-07-12
Report page: switch on Jinja whitespace removal ?


hzeller:
The report page is nicely formatted, but also contains a lot of whitespace that unnecessarily has to be transmitted (the whole page is in the order of 10MiB).

I think Jinja has an option to trim unnecessary whitespace for more compact output.

tgorochowik:
I ran some tests (note that I used only 2 runners to speed up the process, but I am giving the results in percents in hope it reflects running all runners):
* forcing jinja to hide whitespace where it can (substituting `{% %}` with `{%- -%}` etc) saves only about 4%
* jinja is not a html formatter so it still mostly preserves the whitespace from the templates, it can only control whitespace around it's dynamic statemets
* Using obfuscation I was able to save ~39% - the added benefit of this approach is it can also make js and css smaller. I know obfuscation is not what we really want here, but smaller size is a nice benefit of using obfuscation so I think it's worth it

I will create a PR to see what we get for the whole report.

tgorochowik:
When running some tests I noticed that some files for the uhmd-yosys take ~60MB(!) (for example: https://github.com/SymbiFlow/sv-tests/blob/f25bce50e27b2365574f0df735f297ddfa204403/logs/UhdmYosys/generated/black-parrot/bp_processor.sv.log.html )

This is mostly due to dumping a horribly long uhdm tree. I asked @kbieganski to add a switch to the uhdm-yosys integration to be able to disable dumping the tree (this is something we already implement for uhdm-verilator)

kbieganski:
The latest version of Yosys (antmicro/yosys@490d671) does not print the tree by default anymore, so the output size should be reduced significantly.

tgorochowik:
@kbieganski great, thanks!

tgorochowik:
The main page has been optimized #1941, it's now about 300KiB. Let's close this.

PR #956, opened 2020-07-24
[WIP] Obfuscate the output to save space

tgorochowik:
Some experiments for #932 

Do not merge yet. Even if it saves substantial space we need to verify whether everything still works.

### DIFF
--- a/.github/kokoro/run.sh
+++ b/.github/kokoro/run.sh
@@ -78,6 +78,38 @@ echo "----------------------------------------"
 )
 echo "----------------------------------------"
 
+echo
+echo "========================================"
+echo "Obfuscating output"
+echo "----------------------------------------"
+(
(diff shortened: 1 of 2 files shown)


tgorochowik:
Kokoro shows we saved 19132 Bytes (which is only 2%). I will try to create a run which ignores all the log/source files as there is little to save in those and processing them takes most of the time.

tgorochowik:
So for the index + css files alone we save 17856 Bytes (which is 41% of their original size).

I think it's worth considering.

tgorochowik:
Unfortunately the proposed obfuscator seems to completely break the js scripts. Even if we don't touch the js and only change the html files, it looks different (as though not every styling change is applied). In my opinion it makes it unreliable as it is hard to predict the exact effects. Closing.

PR #971, opened 2020-07-31
Rename antmicro-yosys to yosys-sv

tgorochowik:
The origin of the fork is not important from the perspective of
sv-tests.

Rename it to yosys-sv to make it clear that this is a yosys fork capable
of processing more SystemVerilog features and to make its column appear
right next to vanilla yosys.


hzeller:
LGTM

PR #984, opened 2020-08-05
Add fusesoc generator for generating test files

wtatarski:
This PR adds one common fusesoc generator for generating test files according to yaml configuration files in `conf/fusesoc-configs`.


conf/fusesoc-configs/earlgrey.yml
```diff
@@ -0,0 +1,9 @@
+name: earlgrey
+description: Earlgrey design from opentitan
+top_module: top_earlgrey_nexysvideo
+tags: earlgrey uvm
+path: &path third_party/cores/opentitan
+command: ["fusesoc", "--cores-root", *path, "run", "--target=sim", "lowrisc:systems:top_earlgrey:0.1"]
```
tgorochowik, line 6: Could we make it a string and split it to a list in the python script if it's necessary? That would make the config much cleaner.

generators/fusesoc
```diff
@@ -0,0 +1,80 @@
...
+
+for fusesoc_conf_file in fusesoc_conf_files:
+    fusesoc_conf_file = os.path.abspath(
+        os.path.join(conf_dir, "fusesoc-configs", fusesoc_conf_file))
+
+    with open(fusesoc_conf_file, 'r') as stream:
+        try:
+            yml_conf = yaml.safe_load(stream)
+        except yaml.YAMLError as err:
+            print(err)
```
tgorochowik, line 44: Should we also exit here?

generators/fusesoc
```diff
@@ -0,0 +1,80 @@
...
+
+    sources = ''
+    incdirs = ''
+    incs = []
+    command_list = list(yml_conf["command"].split(" "))
```
tgorochowik, line 49: What if one of these keys is unavailable? Maybe we should catch it and report a meaningful message?

generators/fusesoc
```diff
@@ -0,0 +1,98 @@
...
+        yml_conf["top_module"]
+        yml_conf["tags"]
+        yml_conf["path"]
+        yml_conf["command"]
+        yml_conf["conf_file"]
+        yml_conf["test_file"]
+        yml_conf["timeout"]
+    except KeyError as err:
+        print("No key", err, "in file", fusesoc_conf_file)
+        continue
```
tgorochowik, line 62: What I meant is that you could wrap the whole code in try/except so you don't have to have a list of the required keys.
If you want to have a list of the required keys you could have used the `in` keyword and a regular `if` statement.

But let's not spend more time on it. This should work just fine.

PR #986, opened 2020-08-07
Improve test results sorting

### LINKED ISSUE #972: Sorting tool columns: doesn't take partial passing into account
Clicking on the tool column sorts by 'green' and 'red' tests. Partial tests are sorted _after_ these and in an random order. So the sorting column feature is not super useful right now.

If we can have a fraction value attached to each item (0 = all red, 1 = all green, in-between= partial passing), it should be possible to just sort by that.

(I'd probably also not opposed to just remove the sorting feature if that is too complicated. Not sure if it is used)

tgorochowik:
Sort by the percentage of tests passed.

Fixes #972

Expect it to look more or less like this (except this was tested for a single runner):


hzeller:
Ah, very nice output. Thanks Tomasz!

PR #1024, opened 2020-08-23
9.3.4--block_names_par etc: Only check one Variable to appease Verilator

wsnyder:
This modifies these tests to pass on Verilator (which doesn't support multiple parallel fork statements), but still checks what they are intended to test which is the block naming.   Many other tests still check the more complicated fork behavior.

Thanks


tgorochowik (approved):
LGTM, it is enough to test chapter 9.3.4 this test is intended for. Thanks!

PR #1025, opened 2020-08-23
class_tests: Fix missing predeclarations, etc

wsnyder:
This fixes some class tests that are not legal because they do not predeclare referenced classes.

Also class_test_36.sv and class_test_37.sv are removed, IEEE does not allow import directly into classes.


tgorochowik (approved):
LGTM, these `generic` tests are generated from verible and were useful at an early stage, many of them are not completly valid so fixes and removals are welcome, thanks!

PR #1027, opened 2020-08-26
fusesoc: properly order the source files

tgorochowik:
There is a convention saying that we should parse the _pkg.sv files
before the rest of the .sv files. Ibex uses that convention.

CC: @kamilrakoczy


tgorochowik:
This change has been suggested by @kamilrakoczy - we hope it should fix the ibex test for yosys-sv.

piotr-binkowski (approved):
LGTM

tgorochowik:
@hzeller FYI: https://symbiflow.github.io/sv-tests/#YosysSv|ibex|ibex is green with this PR merged

PR #1080, opened 2020-10-02
Make verilator-uhdm run all types of tests

### LINKED ISSUE #1069: uhdm verilator only runs simulation tests
It may be worth considering making uhmd-verilator process most of the tests.

Even if it won't be a full simulation we could at least see if the AST generation process (from the uhdm input) is correct for a given test case.

tgorochowik:
Simply feeding UHDM to Verilator allows us to see if we handle specific
types of nodes, so the Verilator-UHDM integration should run all types
of tests.

Fixes #1069


hzeller (approved):
LGTM

PR #1090, opened 2020-10-08
Update verible linter command name

bluecmd:
Switch to using the current in-use name rather than the deprecated one.

This should remove the warning:
```
*******************************************************************************
verilog_syntax is the deprecated name for verible-verilog-syntax.
executing /home/kbuilder/miniconda/envs/sv-test-env/bin/verible-verilog-syntax instead
*******************************************************************************
```


tgorochowik (approved):
Thanks!

PR #1100, opened 2020-10-17
Remove false test

wsnyder:
This test fails on other simulators and Verilator.

Unbounded is only legal in queues, and other places speced by IEEE, it does not mean "max MSB".

Other tests check $ as a parameter, so just removing this.

Thanks.


piotr-binkowski (approved):
LGTM

PR #1118, opened 2020-10-25
Blacklist yosys values.v which has too many digits to be legal.

wsnyder:
This blacklists a yosys test with illegal numbers.


tgorochowik (approved):
Seems right, thanks!

Maybe it is worth to report it in yosys too?

PR #1128, opened 2020-11-06
Increase default timeout for UVM

dalance:
Now many tests includes UVM, so the default timeout may be too short.


tgorochowik (approved):
We are still within time limit of travis jobs so I guess this is acceptable. Thanks!

tgorochowik:
It looks as though the CI started timing out randomly after merging this (travis has a limit of 50 minutes), we may have to revert this (or try a value somewhere in the middle).

We want to eventually move to using github actions but until this is done we have to do something about the timeouts.

PR #1129, opened 2020-11-06
Fix missing file in ar_src

dalance:
This PR fixes missing files in ar_src.src.


tgorochowik (approved):
We should really convert all the tests to use fusesoc to generate the file lists dynamically (there is a bunch of issues filed for this), but this temporary solution is also welcome, thanks!

PR #1136, opened 2020-11-11
Add :type: simulation

dalance:
This PR adds `:type: simulation` to some tests which should be failed by semantic error.


mithro:
@hzeller - Could you please review?

hzeller:
We should have a 'semantic' category I think.
For now this looks good.

PR #1144, opened 2020-11-16
Remove deadlock from `always` test

### LINKED ISSUE #969: always check has infinite loop
The following code from the test suite has an infinite loop since the always does not have any delay. I would suggest adding a sensitivity to the always (e.g. @(a)) to avoid this infinite loop. Icarus finds and reports this infinite loop and fails instead of allowing the user to run code that will lock up.
```
/*
:name: always
:description: always check
:tags: 9.2.2.1
*/
module always_tb ();
	wire a = 0;
	reg b = 0;
	always
		b = ~a;
endmodule
```
Also keep in mind Icarus is a two stage tool so parsing and elaboration are done in the first stage. It look like a bunch of the other failures reported for Icarus are because it is not just parsing the files, but is also trying to elaborate them. If you are interested I can make suggestions that should allow Icarus to process more of the checks correctly and still test things as expected.

I also have some ideas how to make the ivtest suite a bit more compatible. Many of the skipped tests are because we use a language generation flag to parse the files with the appropriate keywords (language generation). If we also add a compiler directive this should fix the issue when SystemVerilog is assumed.

tgorochowik:
Fixes #969

### DIFF
--- a/tests/chapter-9/9.2.2.1--always.sv
+++ b/tests/chapter-9/9.2.2.1--always.sv
@@ -1,11 +1,9 @@
 /*
 :name: always
 :description: always check
-:tags: 9.2.2.1
+:tags: 9.2.2.1 9.4.1
 */
 module always_tb ();
 	wire a = 0;
-	reg b = 0;
-	always
-		b = ~a;
+	always #5 a = ~a;
 endmodule


tgorochowik:
CC @caryr

hzeller (approved):

piotr-binkowski (approved):
LGTM, but it looks like this makes Icarus runner timeout in CI so we should fix that before merging.

tgorochowik:
Succeeded after a restart. Merging.

PR #1145, opened 2020-11-17
gclk functions require global clocking

wsnyder:
The $*_gclk functions per IEEE require global clocking to be legal.

Also these functions are only legal in assertion contexts in some simulators/formal tools.

Not done but advisable would be a test that checks tools error on gclk without global clocking defined.  I wouldn't check for only assertion contexts as many simulators e.g. Verilator don't care.


piotr-binkowski (approved):
LGTM

PR #1156, opened 2020-11-21
slang runner: Specify top module, run on elaboration tests

MikePopoloski:
The top-module config wasn't being passed to slang, so this fixes that. Also there are a bunch of elaboration tests that I'd like to test slang with, so remove the restriction of just preprocessing and parsing.

### DIFF
--- a/tools/runners/Slang.py
+++ b/tools/runners/Slang.py
@@ -3,7 +3,7 @@ from BaseRunner import BaseRunner
 
 class Slang(BaseRunner):
     def __init__(self):
-        super().__init__("slang", "slang-driver", {"preprocessing", "parsing"})
+        super().__init__("slang", "slang-driver")
 
         self.url = "https://github.com/MikePopoloski/slang"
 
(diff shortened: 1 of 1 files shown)


tools/runners/Slang.py
```diff
@@ -18,6 +18,9 @@ def prepare_run_cb(self, tmp_dir, params):
         # a single compilation unit, so ask slang to do that.
         self.cmd += ['--single-unit']
 
+        if params['top_module'] != '':
```
tgorochowik, line 21: The top module is not explicitly defined for all the tests - because of this we have a helper that tries to guess the top module and use it, you may want to use it too:
https://github.com/SymbiFlow/sv-tests/blob/2aaf56c8ad4204129aa92c5e3fe044ae7e72051b/tools/runners/moore.py#L21-L24

tgorochowik (approved):
LGTM in general! One minor suggestion

tools/runners/Slang.py
MikePopoloski: Thanks, but slang has no problem finding and instantiating the top modules based on the rules laid out in the LRM (all modules that are uninstantiated elsewhere, have no undefaulted parameters, etc).

This change is simply to limit the top modules considered when a test accidentally includes unused modules that should not be checked (which is common for the larger core tests that include unnecessary files).

tgorochowik: Sure, thanks for the explanation

PR #1157, opened 2020-11-21
Fix broken class tests

MikePopoloski:
A bunch of the class tests were invalid for various reasons; most of them because extern methods had no out-of-block implementation provided, or because undeclared types were used.


tgorochowik (approved):
LGTM, thanks!

PR #1159, opened 2020-11-24
Make the default timeout smaller

tgorochowik:
Unfortunately the increased timeout made Travis timeout (50 minute
limit). Until we migrate to github actions the test timeout has to be
smaller.

@dalance we do plan to migrate to github actions so we should be able to restore your change once it is done.


tgorochowik:
Should no longer be needed, the pr with gh actions has been merged

PR #1161, opened 2020-11-25
CI: enable github actions

wtatarski:


wtatarski:
CI run: https://github.com/antmicro/sv-tests/actions/runs/383509585

.github/workflows/install.sh
```diff
@@ -4,7 +4,7 @@ set -x
 set -e
 
 # Add local submodules
-.github/add-local-submodules.sh "${TRAVIS_PULL_REQUEST_SLUG:-$TRAVIS_REPO_SLUG}"
+.github/add-local-submodules.sh "$GITHUB_REPOSITORY_OWNER_SLUG/$GITHUB_REPOSITORY_NAME_SLUG"
```
tgorochowik, line 7: Can't we just use [GITHUB_REPOSITORY](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables) (without the extra action for the env variables?)

tgorochowik (commented):
LGTM in general! some minor comments only

.github/workflows/install.sh
wtatarski: I think it looks better. I used GITHUB_REPOSITORY and removed extra action for env variables.

tgorochowik (approved):

PR #1164, opened 2020-11-26
Generate a fake topmodule for basejump tests

### LINKED ISSUE #1160: Many basejumpSTL tests are bogus
Many of the basejump tests don't work correctly as standalone tests. They reference modules from other files that aren't included in the tests, for example. Most annoyingly, they default all module parameters to "inv" which makes them look like valid top-level modules. All of the tools that correctly follow the SystemVerilog rules for finding top modules will instantiate these and then get stuck in generate loops for millions of iterations. If you look at the top 10 slowest tests for Icarus, Verilator, and slang, they are all basejump STL tests:
```
BaseJumpSTL_bsg_channel_tunnel_wormhole
BaseJumpSTL_bsg_nonsynth_dramsim3
BaseJumpSTL_bsg_mem_multiport
BaseJumpSTL_bsg_nonsynth_ramulator_hbm
BaseJumpSTL_bsg_cache
BaseJumpSTL_bsg_crossbar_control_basic_o_by_i
BaseJumpSTL_bsg_fsb
BaseJumpSTL_bsg_fifo_shift_datapath
BaseJumpSTL_bsg_lru_pseudo_tree_backup
BaseJumpSTL_bsg_cache_to_axi
```

Perhaps this "test suite" should just be removed since it doesn't actually function as a standalone test suite? Alternatively, it seems like you may need a new mode to mark these as "parse-only" or something, to indicate that tools should definitely not try to do type checking or elaboration for them because it's guaranteed to fail. The existing "parse" tag conflates the modes too much to be useful for this I think.

tgorochowik:
Fixes #1160 

@MikePopoloski could you please take a look if this solves the issue for you?

### DIFF
--- a/generators/path_generator
+++ b/generators/path_generator
@@ -6,13 +6,14 @@ import json
 import glob
 
 templ = """/*
-:name: {3}
-:description: Tests imported from {0}
-:files: {2}
-:incdirs: {4}
-:tags: {0}{5}
-{1}
+:name: {name}
+:description: Tests imported from {project}
+:files: {files}
+:incdirs: {incdirs}
+:tags: {project}
+{extra_tags}
 */
+{extra_code}
 """
 
 try:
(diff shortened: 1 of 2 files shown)


generators/path_generator
```diff
@@ -32,20 +33,23 @@ matches = []
...
+        global_extra_tags = []
+        extra_code = ""
         data = json.load(jf)
         name = data['name']
         project = data['project']
         if 'should_fail_because' in data:
-            should_fail_because = ':should_fail_because: ' + data[
-                'should_fail_because']
-        else:
-            should_fail_because = ''
```
MikePopoloski, line 42: Looks like you can remove this variable up above.

MikePopoloski (approved):
Looks good to me, thanks!

generators/path_generator
tgorochowik: Right! removed

MikePopoloski:
It looks like the extra code didn't get included in the basejump tests, so now all of them are failing. I don't know enough about how the generator stuff works to say why it's not taking.

PR #1166, opened 2020-11-26
Add verible verilog kythe extractor runner

mglb:
This adds a runner for `verible-verilog-kythe-extractor`. It detects failure by checking if there were any log messages on stderr, as the tool's exit status is zero even for invalid files.

The verible submodule has been bumped manually, as `verible-verilog-kythe-extractor` changed significantly since the last bump (which happened more than a month ago).

### DIFF
--- a/tools/runners/VeribleExtractor.py
+++ b/tools/runners/VeribleExtractor.py
@@ -1,0 +1,37 @@
+from BaseRunner import BaseRunner
+
+import os
+import shlex
+
+
+class VeribleExtractor(BaseRunner):
+    def __init__(self):
+        super().__init__(
+            "verible_extractor", "verible-verilog-kythe-extractor",
+            {"parsing"})
+
+        self.url = "https://github.com/google/verible"
+
+    def prepare_run_cb(self, tmp_dir, params):
+        src_list_path = os.path.join(tmp_dir, "src_list")
+        script_path = os.path.join(tmp_dir, "run.sh")
+
+        with open(src_list_path, "w") as src_list:
+            print("\n".join(params.get("files", [])), file=src_list)
+
+        inc_dirs = ",".join(params.get("incdirs", []))
+
+        with open(script_path, "w") as script:
+            s = (
+                'log="$({executable}'
+                ' --file_list_root ""'
+                ' --include_dir_paths {inc_dirs}'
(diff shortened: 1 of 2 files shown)


tools/runners/VeribleExtractor.py
```diff
@@ -0,0 +1,66 @@
...
+                pass
+            except ValueError:
+                return ("Invalid OVERRIDE_TEST_TIMEOUTS value", 1)
+
+        try:
+            _, log = proc.communicate(timeout=timeout)
+            # verible-verilog-kythe-extractor always exits with rc=0.
+            # Check error log instead - it should be empty when everything
+            # succeeded.
+            returncode = 0 if not log else 1
```
tgorochowik, line 54: Is this the only reason we have to copy the whole `run_subprocess` method?
Why do we completely remove stdout in this runner? Is there a way we can preserve it?
Maybe if you do have to analyze the output you could do sth like
```
def run_subprocess(self, tmp_dir, params):
    super.run_subprocess(self, tmp_dir, params)
    [ analyze the output here and change the return code ]
```

The thing I want to avoid is having to maintain a very similar code in two places

mglb: `verible-verilog-kythe-extractor` outputs JSON stream on stdout and log messages on stderr. Do we need to store the stream?
`BaseRunner.run_subprocess` assumes log is on stdout and drops stderr, so there's nothing to check.

To avoid code duplication I can modify `BaseRunner.run_subprocess` and move the code for handling process output and status checking (those two lines: https://github.com/SymbiFlow/sv-tests/blob/master/tools/BaseRunner.py#L124) to separate method.

tgorochowik: Ok, I understand the issue now.

Please take a look at the Verilator runner:
https://github.com/SymbiFlow/sv-tests/blob/master/tools/runners/Verilator.py#L67

It generates a helper bash script and then just runs it - you could do something similar.

You could:
* redirect stdout to /dev/null if its not needed
* analyze stderr
* exit the script with the appropriate code

You then will not have to modify `run_subprocess` at all.

Would this work?

mglb: Done

tgorochowik (approved):

PR #1182, opened 2020-11-28
New test from recent IVL_UVM testsuite

svenka3:


tests/chapter-8/8.15--super-default-new.sv
```diff
@@ -0,0 +1,38 @@
...
+    function new ();
+      super.new ();
+    endfunction : new 
+  endclass : uvm_report_object 
+
+endpackage : test_pkg
+
+module m;
+  import test_pkg::*;
+  uvm_object u0;
```
tgorochowik, line 27: Did you mean to use `uvm_report_object`?

tgorochowik (commented):
Thank you for your contribution!

Is there a reason the types are named `uvm_*` even though it is not using uvm, could we use some generic names so it's clear that it doesn't require it?

tests/chapter-8/8.15--super-default-new.sv
svenka3: This is an extract from our IVL_UVM code base that we are working on. I believe this shouldn't be an issue as I don't import uvm_pkg::* here. 

Do you want me to change? I honestly believe this is fine as-is.

Regardss
Srini

tgorochowik: Thank you for your answer!

I think it can stay that way then - btw. are you willing to contribute more of such tests? If these are going to be single snippets like the one you are contributing now then it is fine as it is, but if you are willing to contribute more and it is a part of a larger code base then maybe it would make sense to include your test suite as a whole into the [third party dir](https://github.com/SymbiFlow/sv-tests/tree/master/third_party/tests)?

tgorochowik (approved):

tests/chapter-8/8.15--super-default-new.sv
svenka3: Thanks for the merge, feeling good to have contributed a drop in this ocean :-)

On larger suite - yes, definitely, we are sorting out some of the code clean-up, will consult once we are ready for that.

Also, yes will be keen to add missing low-hanging-fruiths type of tests as well.

Cheers
Srini

PR #1187, opened 2020-12-02
Trigger GHA push action only on the master branch

tgorochowik:
Otherwise it seems to be triggered two times (push and PR) for each
dependabot PR (which pushes to dependabot specific branches without a
fork).


tgorochowik:
Example from: #1176 (note each action being run on two triggers)

PR #1206, opened 2020-12-13
Report test completion date stamp.

wsnyder:
When looking at failing tests on the dashboard it would be helpful to know how long since they last ran.  This adds a date stamp.


tgorochowik (commented):
This is a good idea! Thanks for the PR!
Did you mean to add the timestamp to each test individually? Would a single report-wide timestamp not be enough?

wsnyder:
Individually because a theory of #1207 is what ran later was more likely to fail due to being out-of-space.

tgorochowik (approved):
Makes sense, thanks!

Issue #1146, opened 2020-11-20
Avoid UVM on chapter 18 tests?


wsnyder:
There's many tests in chapter 18 that require UVM.  It seems misleading to users, and discouraging to us tool authors, to have non-green row for e.g. $urandom (18.13.1) for a simulator that supports $urandom but not UVM because 18.13.1--urandom_3.sv fails due to no UVM support.

There's a number of options:

0. Do nothing, it is what it is :)

1. Delete UVM test when there's a corresponding non-UVM test of similar capability.

2. Have the test tag line not list sections other than UVM if UVM is used. e.g. change "tags: 18.13.1 uvm" to "uvm".

3. Other?

If there's no other opinion 2, remove section tag, is my suggestion.  If agreeable I can provide a pull request.

Relevant tests (Some don't have corrsponding non-UVM tests):

```
  18.5--constraint-blocks_1.sv:4::tags: 18.5 uvm
  18.5.1--explicit-external-constraint_2.sv:4::tags: 18.5.1 uvm
  18.5.1--implicit-external-constraint_2.sv:4::tags: 18.5.1 uvm
  18.5.11--static-constraint-blocks_1.sv:4::tags: 18.5.11 uvm
  18.5.12--functions-in-constraint_1.sv:4::tags: 18.5.12 uvm
  18.5.13--constraint-guards_1.sv:4::tags: 18.5.13 uvm
  18.5.14--soft-constraints_1.sv:4::tags: 18.5.14 uvm
  18.5.14.1--soft-constraint-priorities_1.sv:4::tags: 18.5.14.1 uvm
  18.5.14.1--soft-constraint-priorities_3.sv:4::tags: 18.5.14.1 uvm
  18.5.14.1--soft-constraint-priorities_4.sv:4::tags: 18.5.14.1 uvm
  18.5.14.2--discarding-soft-constraints_1.sv:4::tags: 18.5.14.2 uvm
  18.5.14.2--discarding-soft-constraints_3.sv:4::tags: 18.5.14.2 uvm
  18.5.14.2--discarding-soft-constraints_5.sv:4::tags: 18.5.14.2 uvm
  18.5.2--constraint-inheritance_1.sv:4::tags: 18.5.2 uvm
  18.5.2--pure-constraint_1.sv:4::tags: 18.5.2 uvm
  18.5.3--set-membership_1.sv:4::tags: 18.5.3 uvm
[...]

PR #1209, opened 2020-12-13
Update UVM tags to uvm-random (#1146)

wsnyder:
Per comment in #1146 moves tags to all be uvm-random as the current per-section tags in the dashboard don't look good.


tgorochowik (approved):
Sure, let's do this! Thanks

Issue #1162, opened 2020-11-26
9.2.2.1 always test is invalid


caryr:
In the 9.2.2.1 always test both Icarus and Verilator fail because you cannot perform a procedural assignment to a wire. The declaration of "a" needs to be changes to a reg/logic.

caryr:
Fixed in #1210

Issue #1155, opened 2020-11-21
Icarus failing various number_test_* checks


caryr:
As a two stage compiler Icarus verifies that there is a top level module since that is required for generating running code. I noticed that many of the number_test_* files just have a plain parameter definition which is correct for just parsing, but Icarus does both parsing and elaboration. To make these files compatible with Icarus can you add the following to the failing tests listed below.
```
`ifdef __ICARUS__
module dummy;
endmodule
`endif
```
This should allow the tests listed below to pass:
- number_test_0-16
- number_test_24-36
- number_test_42-53
- number_test_59-62
- number_test_64-72

tgorochowik:
@caryr thanks for the report! It should be ok to add the module definition to those tests even without the ifdef.

Would you be willing to create a PR for that?

MikePopoloski:
Not that I disagree with improving the tests, but most tools are able to just print a warning if there is no top module and still compile and check the code. Might be easy enough to modify Icarus to do the same?

caryr:
Yes, I can look at adding parse only support to Icarus. As a minimum it would require a parse only flag since we want to flag a missing module as an error when actually compiling for simulation.

I can look at creating a PR, are there directions for downloading sv-tests and testing using just a single external simulator/test suite? Having that would make it easier for the individual tools providers to test/improve their support.

We are also reworking ivtest so it better supports running all tests as SystemVerilog using an appropriate `begin/end_keywords pair for many of the failing files. This should improve the passing percentage as well as remove a bunch of the black listed files. There are still some others that will require further discussions, but lets get ride of the obvious ones first.

tgorochowik:
> (...) are there directions for downloading sv-tests and testing using just a single external simulator/test suite? Having that would make it easier for the individual tools providers to test/improve their support.

By default sv-tests look for tool executables in PATH, if you do have some extra tools installed, that you just don't want to test feel free to temporarily remove their wrappers from `tools/runners/`, that should do the trick.

I guess you could also modify the Makefile to allow generating the report for just a single tool. Adding an option to run all the tests for a single tool should be straightforward as you could just create a new target in the generator - the part with generating the actual report without trying to rebuild everything may need some more effort. Feel free to send a patch if you come up with something useful :)

caryr:
It looks like the code added in #1164 could also resolve this when the tests listed above are updated to use the new functionality.

tgorochowik:
[...]

PR #1210, opened 2020-12-15
Update wire to logic and number tests to elaborate correctly

caryr:
Fix the elaboration issue discussed in #1162


caryr:
And also the number elaboration issues discussed in #1155

tgorochowik (approved):
LGTM! Thanks for fixing those!

PR #1211, opened 2020-12-15
Do not use the obsolete regress-v10.list in ivtest

caryr:
There likely needs to be some more tweaks added to how the list files are handled but I think this should fix the merge issue in #1205


tgorochowik (approved):
Thanks!

PR #1213, opened 2020-12-16
Add back ivtests that now have a keyword version specified

caryr:

### DIFF
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -39,118 +39,63 @@ ivtest_exclude = [
 ]
 
 ivtest_blacklist = [
-    'andnot1',  # ref is keyword
-    'automatic_error11',  # local is keyword
-    'automatic_error12',  # local is keyword
-    'automatic_error13',  # local is keyword
     'binary_nand',  # ~& is not binary operator
     'binary_nor',  # ~| is not binary operator
     'bool1',  # reg can't have type
-    'br930',  # type is keyword
-    'br937',  # expect is keyword
     'br943_944_iv.sv',  # missing submodule
     'br974c',  # reg and logic is exclusive
     'br985_iv.sv',  # missing submodule
     'br988',  # generate_item can't have begin/end
-    'br_gh115',  # bit is keyword
     'br_gh72b',  # should_fail test
     'br_gh99c_iv.sv',  # vams only
     'ca_time_iv',  # IV only $simtime
     'case3',  # '@ *' is invalid event_control ( '@*' or '@ (*)' is valid )
-    'case5',  # bit is keyword
-    'case5synfail',  # bit is keyword
-    'case6',  # bit is keyword
-    'case7',  # bit is keyword
     'cast_int_iv.sv',  # vams only
     'compare_bool_reg',  # reg can't have type
-    'condit1',  # bit is keyword
     'constfunc4_iv.sv',  # vams only
     'constfunc6_iv',  # IV only $abs, $min, $max
     'constfunc8',  # reg can't have type
-    'delay_var',  # int is keyword
     'deposit_iv',  # IV only $deposit
     'deposit_wire_iv',  # IV only $deposit
     'display_bug',  # [2] is invalid range
-    'drive_strength3',  # bit is keyword
     'elsif_test',  # ifdef must have text_macro_identifier
     'enum_ports',  # reg can't have type
-    'fifo',  # expect is keyword
-    'fileio',  # var is keyword
-    'fread',  # string is keyword
-    'gencrc',  # bit is keyword
-    'generate_multi_loop',  # bit is keyword
     'ivlh_textio_iv',  # IV only $ivlh_file_open
     'macro_str_esc',  # `` is invalid macro usage
     'macro_with_args',  # `` is invalid macro usage
-    'memsynth3',  # bit is keyword
-    'nb_assign',  # var is keyword
-    'nb_ec_pv2',  # bit is keyword
-    'no_if_statement',  # var is keyword
     'nonpolymorphicabs_iv',  # IV only AMS $abs
     'plus_arg_string_iv',  # IV only $finish_and_return
-    'pow_ca_signed',  # expect is keyword
-    'pow_ca_unsigned',  # expect is keyword
-    'pow_reg_signed',  # expect is keyword
-    'pow_reg_unsigned',  # expect is keyword
     'pr478',  # `protect is not valid directive
-    'pr529',  # bit is keyword
     'pr622',  # `` is invalid macro usage
     'pr639',  # `` is invalid macro usage
     'pr841_iv.sv',  # unresolved wires
     'pr1467825',  # `suppress_faults is not valid directive
-    'pr1478121',  # do is keyword
     'pr1494799_iv',  # IV only $is_signed
-    'pr1676071',  # bit is keyword
-    'pr1698820',  # var is keyword
     'pr1716276',  # empty parameter is invalid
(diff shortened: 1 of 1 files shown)


tgorochowik:
@wsnyder as you were the one to extend the current blacklist (which I think this PR mostly reverts) - would you mind taking a quick look at the tests to see if your concerns are indeed already resovled?

wsnyder:
Most are ok but these are still broken IMO.

fileio_iv.sv
* Not legal - $fopenr/$fopenw are not IEEE

fread_iv.sv
* Not legal - $fread from invalid variable

memsynth3_iv.sv
* Not legal - need to add language_keywords

pr1978358_iv.sv
pr1978358b_iv.sv
pr1978358c_iv.sv
pr1978358d_iv.sv
sel_rval_bit_ob_iv.sv
* Not legal or at least not clean - select out of bounds

caryr:
I will put these back in the black list for now. Not sure how my force SystemVerilog testing missed the memsynth3 test. I will need to investigate and resolve. Some of these will need more conversations.

caryr:
Where is the best place to discuss this. For example I believe the selects in the various pr1978358 tests are valid SystemVerilog. From 1800-2012 section 11.5.1:

> A part-select that addresses a range of bits that are completely out of the address bounds of the vector,
> packed array, packed structure, parameter or concatenation, or a part-select that is x or z shall yield the value
> x when read and shall have no effect on the data stored when written. Part-selects that are partially out of
> range shall, when read, return x for the bits that are out of range and shall, when written, only affect the bits
> that are in range.

A similar paragraph exists for bit selects.

wsnyder:
> 
> 
> Where is the best place to discuss this? For example I believe the selects in the various pr1978358 tests are valid SystemVerilog. From 1800-2012 section 11.5.1:
> 
> > A part-select that addresses a range of bits that are completely out of the address bounds of the vector,
> > packed array, packed structure, parameter or concatenation, or a part-select that is x or z shall yield the value
> > x when read and shall have no effect on the data stored when written. Part-selects that are partially out of
> > range shall, when read, return x for the bits that are out of range and shall, when written, only affect the bits
> > that are in range.
> 
> A similar paragraph exists for bit selects.

Ah, but: "NOTE 1-Part-select indices that evaluate to x or z may be flagged as a compile time error.

Which is why for those tests I said "Not legal or at least not clean".  I don't believe sv-tests has a way at present to indicate that tests are legal code, but a linter tool might fail because there's something that it is flagging as dubious coding.  If there's a way to flag these tests in such a way they can be merged.

MikePopoloski:
Yeah, it seems vastly preferable to be notified of your coding errors at compile time where possible. If icarus can't do that, fine, the spec leaves open the option to detect at sim time, but that should not be baked into the test suite as "failing" tests for tools with better compile-time checking.

caryr:
Oh, I agree this should be flagged and in Icarus we have this as part of our normal -Wall warning group. So these are not invalid per the standard since it specifically uses "may" it's just there are different opinions if this should be a warning or an error.

caryr:
Also note that this error out exception applies only to part selects and not bit selects. We chose to treat both uniformly which is generate a warning. I will agree having an option to error out for these out of range selects may be a good choice, but we have tried to generate only meaningful warning so recommend using -Wall for Icarus and keeping the warnings clean.

caryr:
I checked sel_rval_bit_ob and it appears to only use out of range bit selects so I believe the "may error" clause doe not apply to them. What is strange is @wsnyder did not complain about sel_rval_part_ob which does have out of range part selects. Maybe this is an artifact of the funky ivtest list handling in sv-tests.

MikePopoloski:
Note that there are many more tests that I think should be blacklisted for similar reasons but aren't yet. I wouldn't take that to mean they are all ok, it's more likely we just haven't had time to look through the hundreds of tests and find them all.

caryr:
Okay, my goal over the next few weeks is to get things like this addressed and configured correctly. As a minimum I would like to get iverilog clean for ivtest and then will see if I can identify all the non-standard behaviour in ivtest/iverilog so it can be excluded or updated to work correctly. Much of this is related to getting the test list generated correctly so the known extensions (VHDL, VAMS, etc.) are excluded correctly.

wsnyder:
Note I blacklisted based on what Verilator failed and seemed obvious, e.g. sel_rval_part_ob is a case where Verilator can't run because of other unsupported stuff.

caryr:
Makes sense. What's your take on sel_rval_bit_ob. That seems like something Verilator is incorrectly claiming is invalid.

wsnyder:
sel_rval_bit_ob fails on Verilator because of the 1'bx's, not because of the OOB; Verilator warns abouut the little endian usages (line 16,18,20,22,24) and out of range (lines 45/53) but in this case are just warnings not errors.

caryr:
What is the justification for the 1'bx failing in a bit select? Is see where a part select may error out, but cannot find where a bit select is allowed to error out.

In Icarus since out of range and undefined select indexes have a clearly defined operation we have chosen to implement this functionality and provide warnings when this can be detected at compile time. This provides consistent operation for constant and variable (indexed) selects.

Also I just noticed things have changed or I was not remembering correctly, but out of range bit selects may also be a compile time error.

MikePopoloski:
Such selects are clearly bogus; it's fine for Icarus to warn instead of error, we're just saying it should also be fine for a tool to error instead of warn. These tests should not be included in a general test suite for all SystemVerilog tools for that reason.

wsnyder:
Sorry, I wasn't being clear, I think it is probably legal for sel_rval_bit_ob to pass.  Verilator specifically dies on it because Verilator is a 2-state simulator and so throws an UNSUPPORTED error on the 1'bx, this doesn't affect to icarus in any way.  I mentioned that only because I wasn't analyzing sel_rval_bit_ob validity further because of Verilator's UNSUPPORTED, not because I thought the 1'bx was illegal.

caryr:
@MikePopoloski I completely agree and understand. I just want exclusions like this to have clear documentation. I am locally adding better comments for the exclusion and may tweak thing more before I commit. The ideal is for ivtest to be 100% compliant and the exclusions are only for thing where a choice in implementation is allowed by the standard. There will be more discussions like this going forward to reach that goal.

@wsnyder  Thanks for clarifying. I'll double check the exact usage and believe this will ultimately be excluded because of the may error for constant out of range bit select. I sometimes forget Verilator is only two state. Your life is so much simpler!

I believe this PR is currently clean for merging.

caryr:
Any idea why this would be happening when all this patch is changing is ivtest related files?

> ERROR   | Unable to test Yosys using generated/basejump/BaseJumpSTL_bsg_launch_sync_sync.sv: psutil.NoSuchProcess process no longer exists (pid=2409)
> make: *** [out//logs/Yosys/generated/basejump/BaseJumpSTL_bsg_launch_sync_sync.sv.log] Error 1
> Makefile:146: recipe for target 'out//logs/Yosys/generated/basejump/BaseJumpSTL_bsg_launch_sync_sync.sv.log' failed
> make: *** Waiting for unfinished jobs....
> Error: Process completed with exit code 2.

It was clean before I pushed the latest few commits. Seems like a system issue so is there a way to retrigger the build?

caryr:
I added a comment to retrigger the build.

caryr:
@wsnyder all the tests you requested have been added back to the exclusion list. Can you please verify everything is good from your end.

wsnyder (approved):

tgorochowik:
Thank you all!

PR #1226, opened 2020-12-19
Remove ivtest blacklist items already excluded

caryr:
This removes a bunch of items that were blacklisted, but with the recent changes I made to the exclude list generation they should not longer even be available for removal.

### DIFF
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -50,75 +50,47 @@ ivtest_exclude = [
 ]
 
 ivtest_blacklist = [
-    'binary_nand',  # ~& is not binary operator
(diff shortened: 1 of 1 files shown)


caryr:
Can this get merged? Is there anything else you are waiting for?

tgorochowik (approved):
Thanks for your work @caryr - I am sorry for the delay.

In the future if you mention any _recent changes_ it would be cool if you could point to the changes and explain why what your saying is okay to do - this way it would be easier for people to verify your PR.

For now I quickly went through the updated lists and they indeed seem to exclude the tests you're removing from the blacklist anyway. Thanks!

PR #1229, opened 2020-12-25
Fix test type

dalance:
The test type should be `simulation` because this test detects semantic error.


tgorochowik (commented):
Thanks for the PR! I understand the issue you are trying to solve but I am not sure this is the correct way to do that.

The simulators will most probably fail this test as there is no top module.

Maybe the correct way to handle that would be adding an elaboration-specific type and marking the appropriate tools as capable?

tgorochowik:
We've added an 'elaboration' type since this PR has been created. I've rebased your code and changed the type to 'elaboration'.

PR #1230, opened 2020-12-25
Fix syntax error of inside expression

dalance:
inside_expression can take a expression at the left side of `inside`.
Therefore `b1, b2` can't be taken.

```
inside_expression ::= expression inside { open_range_list }
```


tgorochowik (approved):
Looks correct, thanks!

PR #1240, opened 2020-12-30
Conda channels fix

lpawelcz:
This PR is fixing old verilator version in conda environment. Basically, conda was getting this package from channel `conda-forge` instead of `symbiflow`. I forced all packages listed as jobs to be fetched from channel symbiflow. Additionally I also had to change `sed` command in `install.sh` to take into account new dependency format while uncommenting packages.

Creating this as draft in order to make sure that CI is ok.


tgorochowik (approved):
LGTM!

PR #1243, opened 2021-01-01
Update class tests to have a module definition

caryr:
Since iverilog does full elaboration it needs to have a top level module so I added a dummy module to all the class tests. This is similar to #1210 from a few weeks ago for the number tests.


tgorochowik (approved):
This is acceptable, thanks!

Issue #1241, opened 2020-12-30
CI runs are taking significantly longer to complete


caryr:
It looks like the install phase for each of the tools is taking significantly longer. This is almost doubling the time for some of the tests and is a huge waste of resources. For example many of the test were previously installed in about 5 minutes, but they are now in the half an hour range.

caryr:
Just to put this in perspective, this is over five hours of compute time per CI run so for the typical pull and then integrate that's over ten hours per pull request.

mithro:
I think this is likely to be because https://github.com/SymbiFlow/sv-tests/issues/1179

caryr:
I don't remember when this started being slower, but I think it was after #1179. I could be related to my cloning with a depth of 1 which is related to #1179. Though I think this was still okay after I pushed my first change.

caryr:
FYI #1226 was the last PR to run/be integrated with the shorter install times so that does point at the depth of 1 clone as the likely culprit. Not sure what needs to be done to fix this and allow shallow clones. What is odd is I did do an earlier commit, but that may have been before I initialized the submodules so maybe that's a clue. Looks like a git magician is needed here :-).

mithro:
After merging #1242 and removing the gh-pages branch from this repository. This bug should hopefully be fixed.

mithro:
Hopefully the CI run at https://github.com/SymbiFlow/sv-tests/actions/runs/455335796 will show this is now fixed...

caryr:
I took a quick look at the link and the Install phase is still around half an hour.

mithro:
@caryr You appear to be correct. I took a closer look at the run time and it seems to be mostly caused by the [`add-local-submodule.sh`](https://github.com/SymbiFlow/sv-tests/blob/master/.github/add-local-submodules.sh) script doing the checkout of all the submodules under `./third_party/cores/black-parrot/external/` particularly `./third_party/cores/black-parrot/external/riscv-gnu-toolchain`.

tgorochowik:
@mithro I think we could also consider getting rid of all the tool repos from the `third_party` dir. Originally we used that to build all the tools before testing and later kept them to make it possible to run everything locally without conda.

However I don't think anybody uses that anymore. It is much simpler to use the conda packages in the CI and locally everybody seems to be interested in testing a single tool from their own sources (which is understandable)

tgorochowik:
[...]

PR #1244, opened 2021-01-01
Use treeless clones for submodules

mithro:
Using the tips found in https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ this changes all the submodules to use "treeless" clones.

> * `git clone --filter=tree:0 <url>` creates a treeless clone. These clones download all reachable commits while fetching trees and blobs on-demand. These clones are best for build environments where the repository will be deleted after a single build, but you still need access to commit history.

This makes the amount of data downloaded significantly smaller. This reduces the time taken to clone all the submodules by about a third and helping with https://github.com/SymbiFlow/sv-tests/issues/1241 a bit (30m -> 17m).

### DIFF
--- a/.github/add-local-submodule-inner.sh
+++ b/.github/add-local-submodule-inner.sh
@@ -50,12 +50,14 @@ fi
 # If submodule doesn't exist, clone directly from the users repo
 if [ ! -e $SM_PATH/.git ]; then
 	echo "Cloning '$SM_PATH' from repo '$ORIGIN_URL'"
-	git clone $ORIGIN_URL $SM_PATH --origin origin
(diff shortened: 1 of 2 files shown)


mithro:
A longer term fix is to finish my Python version of the `add-local-submodules.sh` script. I'm **sure** I had started it, but I totally can not find any reference to now...

mithro:
FYI - I talk about having a Python version of the above script in https://github.com/enjoy-digital/litex/issues/394 too...

mithro:
https://github.com/mithro/sv-tests/actions/runs/456749951 shows the faster builds.

PR #1266, opened 2021-01-10
Fix arguments.sv broken test condition

wsnyder:
Signed-off-by: Wilson Snyder <wsnyder@wsnyder.org>


tgorochowik (approved):
Right, thanks!

PR #1270, opened 2021-01-12
VeribleExtractor: detect segfaults and ignore redefined macros

mglb:
Two fixes for how results are interpreted:
* Segfaults are reported as failures
* `Re-defining macro` warnings do not cause test failure

### DIFF
--- a/tools/runners/VeribleExtractor.py
+++ b/tools/runners/VeribleExtractor.py
@@ -33,16 +33,28 @@ class VeribleExtractor(BaseRunner):
         inc_dirs = ",".join(params.get("incdirs", []))
 
         with open(script_path, "w") as script:
-            s = (
-                'log="$({executable}'
+            command = (
+                '{executable}'
                 ' --file_list_root ""'
                 ' --include_dir_paths {inc_dirs}'
-                ' --file_list_path {src_list_path}'
-                ' 2>&1 1>/dev/null)"\n'
-                'if [ -n "$log" ]; then echo "$log"; exit 1; fi\n').format(
+                ' --file_list_path {src_list_path}').format(
                     executable=self.executable,
                     inc_dirs=shlex.quote(inc_dirs),
                     src_list_path=shlex.quote(src_list_path))
+            s = (
+                'echo "#" {command_str}\n'
+                'log="$( {command} 2>&1 1>/dev/null )"\n'
+                'rc=$?\n'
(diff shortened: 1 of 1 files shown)


tools/runners/VeribleExtractor.py
```diff
@@ -33,16 +33,27 @@ def prepare_run_cb(self, tmp_dir, params):
...
                 ' --file_list_root ""'
                 ' --include_dir_paths {inc_dirs}'
-                ' --file_list_path {src_list_path}'
-                ' 2>&1 1>/dev/null)"\n'
-                'if [ -n "$log" ]; then echo "$log"; exit 1; fi\n').format(
+                ' --file_list_path {src_list_path}').format(
                     executable=self.executable,
                     inc_dirs=shlex.quote(inc_dirs),
                     src_list_path=shlex.quote(src_list_path))
+            s = (
```
hzeller, line 44: Maybe add a comment that describes what is happening  - that we use the log output as a proxy information that something failed unless it is talking about re-defining a macro.

tools/runners/VeribleExtractor.py
```diff
@@ -33,16 +33,27 @@ def prepare_run_cb(self, tmp_dir, params):
...
+                'log="$( {command} 2>&1 1>/dev/null )"\n'
+                'rc=$?\n'
+                'echo "stderr:"\n'
+                'echo "$log"\n'
+                'if [ $rc -eq 0 ]; then\n'
+                '    ! echo "$log" | grep -v -e "^I .*] Re-defining macro.*" -e "^$" -q\n'
+                '    rc=$?\n'
+                'fi\n'
+                'exit $rc\n').format(
+                    command_str=shlex.quote(command),
```
hzeller, line 55: Looks like the Python [formatter was not happy](https://github.com/SymbiFlow/sv-tests/pull/1270/checks?check_run_id=1688950278#step:6:1) here.

hzeller (changes requested):
LGTM, just two little things.

tools/runners/VeribleExtractor.py
mglb: Fixed

tools/runners/VeribleExtractor.py
mglb: Fixed

tgorochowik (approved):
LGTM, let's merge it

Issue #937, opened 2020-07-14
Automatically determine file list for Ariane test


piotr-binkowski:
Currently Ariane test uses hardcoded list of files that can become outdated after Ariane update (#936)
It would be best to use FuseSoC to prepare a list of files just like we did in case of Earlgrey (#875) since Ariane [seems to have FuseSoC support](https://github.com/openhwgroup/cva6/blob/2d814452094b3e20897e2362771d70024a610651/ariane.core)

tgorochowik:
Done in #1275

PR #1275, opened 2021-01-15
Use file list for Ariane test from cva6 submodule

joannabrozek:
Ref: #937

Ariane repository has added the fusesoc support, but it isn't maintained and the current file list is deprecated. There is another file list available, which supports OpenPiton and it seems to be up-to-date. This list was used to create this PR to replace the hardcoded list of files for Ariane test.


tgorochowik (approved):
LGTM, thanks!

Issue #1271, opened 2021-01-12
Invalid ivtests that should be blacklisted?


MikePopoloski:
I went through failing ivtests for slang. Some of them I believe are failing because the tests themselves are invalid. They fall into the following groups:

### Invalid net types
A bunch of tests use 'real' or 2-state net types, which are not allowed in SystemVerilog. Commercial tools fail on this as well. There are many, some examples:
* implicit_cast12_iv
* implicit_cast13_iv
* ca_time_iv
* br_gh99e_iv

### Mod of real number
This is not allowed in SystemVerilog. Commercial tools fail on this as well. There are many, some examples:
* cfunc_assign_op_real_iv
* cfunc_assign_op_mixed_iv
* mixed_type_div_mod_iv
* pr1528093_iv

### Variables used before they are declared
This is not allowed in SystemVerilog. Some commercial tools correctly fail on this, some do not. Some examples:
* pr1936363_iv
* pr1648365_iv

### Use of optional system functions
This is arguable, but the spec clearly says tools may not have these functions. It doesn't make sense to me to have them in the test suite.

### Use of unknown system functions
Examples: $deposit, $abstime are not specified anywhere in SystemVerilog.
* sysargs_iv
* pr2590274a_iv

### Invalid use of system functions
There are a bunch of tests that pass the wrong number of arguments to system functions, or arguments with incorrect types. Perhaps these are supposed to be marked as should_fail ?
* sf_countbits_fail.iv
* sf_countones_fail.iv
* sf_isunknown_fail.iv
* sf_onehot0_fail.iv
* sf_onehot_fail.iv
* scaninvalid.iv
* readmeminvalid.iv
* freaderror_iv
* pr2800985b_iv

### Miscellaneous
* pr1916261_iv  - uses a replication constant of 0, which is not allowed
* param_test3_iv - provides parameter assignments to a module that does not have any parameter declarations
* elsif_test_iv - has multiple \`else directives for the same \`if directive (this is probably just a typo in the test? not sure why Icarus doesn't catch it)
* comment1_iv - has nested block comments, which are not allowed

@caryr can you take a look?

caryr:
[...]

PR #1286, opened 2021-01-18
Update ivtest with code cleanup

caryr:

### DIFF
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -107,6 +107,51 @@ ivtest_file_exclude = [
     'countdrivers4',
     'countdrivers5',
     'warn_opt_sys_tf',
+    # Icarus checks that always blocks have delay constructs to prevent
+    # infinite loops. For now exclude these. I would ideally like to make
+    # these marked as fails for Icarus and and correctly parsed for others.
+    'always4A',
+    'always4B',
+    'always311A',
+    'always311B',
+    'always312A',
+    'always312B',
+    'always312C',
+    'always312D',
+    'always312E',
+    'always312F',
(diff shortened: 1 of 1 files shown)


tgorochowik:
@caryr would you mind providing a short summary of the changes in your PRs regarding the ivtests updates?

Your work is very much appreciated and having that sort of summary would allow us to verify those changes much quicker

caryr:
I updated ivtest and added/removed some items in the exclusion list for ivtest to make it more compatible given the requests in #1271. This is 100% focused on ivtest and what part of it is used.

caryr:
Are you requesting a list of changes to the actual ivtest repo? Because checking in the two commits should have made it obvious the simple changes made inside sv-tests. You can assume I will not remove anything from the exclusion list until I expect it to be handled correctly in ivtest.

caryr:
I just pushed some more changes: This reworks the following:

Moves some tests that are not implemented in Icarus yet from the CE (compile error) to normal so they will pass in sv-tests for tools that do support the functionality.

For now exclude the tests that check for always processes having delays. Eventually this should be marked as passing for other tools that do not check this and fail for Icarus

We have a similar check for unexpected delays added to always_* process and these have also been excluded.

Eventually I would like to make the above two know which environment was being used and switch from CE to normal depending on which tool is processing them. This will require some larger changes to sv-tests since the ivtest list needs to be generated more dynamically.

caryr:
I also adds the ability to skip test marked NI (Not Implemented)

tgorochowik (approved):
Sounds good, thanks

PR #1298, opened 2021-01-27
Update ivtest and exclude list

caryr:

Update to the latest ivtest repo and add a few more tests to the exclude list that are defacto standard/supported by commercial simulators.


caryr:
FYI after talking with @wsnyder in email regarding two of the excluded tests we agreed that one was incorrectly excluded and the other I made a minor tweak to satisfy his concern.

tgorochowik (approved):
Thanks!

PR #1299, opened 2021-01-27
Bumping BlackParrot and updating generator

dpetrisko:
Title, fixes some of the Verilator tests


tgorochowik (approved):
Thanks for the update!

PR #1301, opened 2021-01-27
Add tests for triggering and waiting for named events

MegabytePhreak:
These are described in chapter 15.5 of the LRM.

The tests cover direct and hierachical references to events within
modules, as well as direct references within a class.

### DIFF
--- a/conf/lrm.conf
+++ b/conf/lrm.conf
@@ -377,6 +377,7 @@ ivtest	Tests imported from ivtest	https://github.com/steveicarus/ivtest
 13.4.4	Background processes spawned by function calls
 14.3	Clocking block declaration
 15.4	Mailboxes
+15.5	Named Events
 16.2	Overview
 16.4	Deferred assertions
 16.7	Sequences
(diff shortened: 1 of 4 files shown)


caryr:
Should the normal and non-blocking event checks be in separate tests?

MegabytePhreak:
I'm open to the idea. I'm not sure what the project preference is. They both come under the same section in the LRM. 
I've honestly never used the non-blocking form myself. 

I actually found an interesting hole in the spec while writing this test. Namely you can put an event in a class, but can't refer to it using this.some_event for either triggering or waiting. Multiple commercial tools will accept this, but reading the LRM it doesn't seem to be permitted, since this is a reserved keyword and not an identifier.

tgorochowik (commented):
Thank you for your contribution!

Looks good to me in general, I agree with @caryr that it would be nice to create two separate tests for blocking and non-blocking triggers - would you mind splitting that to two tests?

tgorochowik (approved):
Since these are not a proper simulation tests but rather tests that verify whether the operators can be parsed in various contexts I think the changes are enough.

LGTM, thanks for the contribution!

PR #1317, opened 2021-02-03
Introduce report analyzer

lpawelcz:
This PR introduces report analyzer script that takes 2 sv-tests CSV reports and compares them. It returns JSON file with given format:
```
{
    "comparable_tools": {
        "YosysSv": {
            "new_passes": [
                "BrokenTest_1",
                "BrokenTest_2"
            ],
            "new_failures": [
                "FixedTest_1",
                "FixedTest_2"
            ],
            "added": [
                "RemovedTest_1",
                "RemovedTest_2"
            ],
            "removed": [
                "AddedTest_2",
                "AddedTest_1"
            ],
            "summary": {
                "new_failures": 2,
                "new_passes": 2,
                "added": 2,
                "removed": 2,
                "not_affected": 96
            }
        }
    },
    "added_tools": [
        "RemovedTool_2",
        "RemovedTool_1"
    ],
    "removed_tools": [
        "AddedTool_2",
        "AddedTool_1"
    ]
}
```
It also outputs small summaries for each tool to stdout and log file (can be disabled with `-q`)
There is added CI integration for this script: each CI job after running tests takes generated `report.csv` and compares it against latest `report.csv` from https://symbiflow.github.io/sv-tests-results/. Result of the comparison is then uploaded as an artifact for each job.


.github/workflows/report_analyzer.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 The SymbiFlow Authors.
```
tgorochowik, line 4: 2021

tgorochowik (commented):
LGTM in general.

Two things to do:
* Move the tool to the `tools` directory, I think it is useful outside of CI too
* Make sure to modify the `make format` target to take that file into account too (it could possibly work out of the box  if you move it to the dir mentioned in the previous bullet)

tgorochowik:
Maybe it would also make sense to run this in kokoro as well? This way we would have a full summary in one place.

lpawelcz:
changes in v2:
* updated year in copyrights
* moved `report_analyzer.py` into `tools` directory
* formatted the code with `make format`
* CI: report from current master is downloaded to `out/report/base_report.csv`
* CI: summary generated to `out/report/changes_summary.json`
* Added generating summary in kokoro script

.github/workflows/report_analyzer.py
lpawelcz: fixed

lpawelcz:
Changes in v3:

- removed logging feature in favor of generating markdown table with summary for each tool in report (`pytablewriter` dependency) and outputting it to file
- updated kokoro script to handle 2 output files (summary JSON and MD table)

Additionally:

- removed summary generation in each specific tool job
- each tool job exports only its `report.csv` as an artifact
- introduced `Summary` job that downloads generated previously reports, then it concatenates them and generates one big summary for all tools. Then it deletes the artifacts with tool reports as they are no longer needed. Summary results (JSON and MD table), together with mainline `report.csv` and compared one are uploaded as an artifact
- introduced `Comment` job that takes output markdown table from `report_analyzer` and puts it in the PR comment. When similar (first lines are identical) summary comment is already present in PR, the bot deletes it and replaces it with a new one

lpawelcz:
Changes in v4:
* Moved `Comment` job to separate workflow triggered by event `workflow_run`
* Modified download-artifacts step for downloading tests summary
* Final summary artifact has additional file with pull request number for identification purposes

In v3, putting comments on PR from forked repositories was not working due to github restrictions for workflow tokens triggered by event `pull_request`. In this version we introduce safe method for running tests on code from PR and putting the comment with tests summary.
Tokens in workflows triggered by `workflow_run` event have increased privileges so they are capable of putting the comments on PRs. The important thing about those workflows is that they are executed against base branch of the PR (for example `master`) so that no potential malicious code from PR will be executed with RW permission. By using this event we are able to isolate the `Comment` job from rest of the CI that is operating on untrusted code from PR.
`comment-pr` workflow is triggered on each completion of `sv-tests-ci` but it actually executes only for PRs and only if `sv-tests-ci` succeeded.
We had to change the method for downloading the artifacts since the `comment-pr` needs to download artifacts from other workflow (`sv-tests-ci`). Such case is not supported in standard `actions/download-artifact@v2` so it was replaced by custom script.
In order to put a comment, a PR number is required.  The `workflow_run` event lacks that information in its context so it is required to pass it together with the artifacts form previous workflow so that comments can be placed in correct PRs.

Having in mind that workflows triggered on `workflow_run` execute on base of the PR, we know that this workflow won't run here in this PR (as there is no such workflow defined in `SymbiFlow:master`). Commenting feature will work on next PRs only after merging this one into master.

Solution based on: https://securitylab.github.com/research/github-actions-preventing-pwn-requests

tgorochowik (approved):
LGTM!

PR #1329, opened 2021-02-11
Test report analyzer: temporary PR

lpawelcz:
This PR is for testing changes introduced by https://github.com/SymbiFlow/sv-tests/pull/1317
CI run in this PR should execute workflow `sv-test-ci` and after that, second workflow `comment-pr` should be triggered. 
We expect that this workflow will put a table with comparison of test results between current results from https://symbiflow.github.io/sv-tests-results/ and this PR.

PR will be closed after confirming that commenting feature is working properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# SystemVerilog Tester
+# SystemVerilog Tester - testing report analyzer
 
 [![License](https://img.shields.io/github/license/SymbiFlow/sv-tests.svg)](https://github.com/SymbiFlow/sv-tests/blob/master/LICENSE)
 [![Build Status](https://github.com/symbiflow/sv-tests/workflows/sv-tests-ci/badge.svg)](https://github.com/SymbiFlow/sv-tests/actions)


tgorochowik:
It works 🎉

CC @hzeller

hzeller:
:raised_hands: Yay, this is really useful!

Now the question I guess is why there were any changes at all, as this was just a README change and why are some tests in some tools removed ?

Do some of these tests are close to timeout or have random crashes so that could explain these differences ? If so, we might need to dig down and find where the flaky invocations are. There is for instance the `AllowedTimeout` column; if we compare that with the `TimeWall` we can find timeouts.

tgorochowik:
I think that most of these changes may be caused by new conda packages versions which are built independently from this repository. The changes in Verilator and Slang are probably mostly caused by this.

But the smaller numbers could probably mean that there are some flaky tests, timing out is indeed a plausible reason.

What is even weirder though is why some of the tools ran 15 fewer tests than earlier. Even the generated tests come from submodules so it shouldn't have changed.

mithro:
@tgorochowik -- We should setup a github action / kokoro run which runs the tests on a regular schedule and checks the output changes?

tgorochowik:
The report is currently uploaded by kokoro so I think it would be preferable to setup periodical (daily?) builds using it. Can this be done?

I know it is fairly easy to setup a cron job for gha but that would mean we would have to move to whole report generation to gha too

mithro:
Yes, I can set up period builds on Kokoro.

mithro:
@tgorochowik - If you add a `periodic.cfg` file in [`.github/kokoro`](https://github.com/SymbiFlow/sv-tests/tree/master/.github/kokoro) (see the [`continuous.cfg`](https://github.com/SymbiFlow/sv-tests/blob/master/.github/kokoro/continuous.cfg) as an example) then it should run every 6 hours.

PR #1331, opened 2021-02-12
Fix module path for UhdmYosys tool

joannabrozek:
The test for ibex core and UhdmYosys tool finishes with the following error:
```ERROR: Module `\work_ibex_core' not found!```
This commit fixes it.


tgorochowik (approved):
LGTM

PR #1342, opened 2021-02-15
summary: simplify adding header

joannabrozek:
This commit fixes error about broken pipe, for example:
https://github.com/SymbiFlow/sv-tests/runs/1903030447?check_suite_focus=true#step:6:2578


tgorochowik (approved):
LGTM

PR #1370, opened 2021-03-04
Make sure ivtests generates tests with unique names

RRozak:
There were a few tests that had the same name and were on the different lists. The generator was assuming that each test has distinct name and iterated through the list of lists of tests in undeterministic order. The tests with the same names were overwriting one another.
I made the iteration deterministic and tests names distinct by adding a list names as their prefixes.


tgorochowik (approved):
LGTM

PR #1379, opened 2021-03-06
Add type() operator comparison test

wsnyder:
This tests type() comparisons, which seem to blow up several tools.


tgorochowik (approved):
I see that the CI fails because of some changes in the tree-sitter runner, but that seems to be a separate problem.

Note that the test type is not specified so only the syntax will be checked for it (no attempts to run the simulation will be done.

tgorochowik:
@drom the runner for tree-sitter-verilog seems to fail with the following error:
`Incompatible Language version 12. Must be between 13 and 13`
Did something change in the latest version of your library? Should we adjust our runners?

wsnyder:
Would you prefer I add the tag so it simulaltes?

tgorochowik:
No, it's ok as it is, just making sure it was your intention.

drom:
> @drom the runner for tree-sitter-verilog seems to fail with the following error:
> `Incompatible Language version 12. Must be between 13 and 13`
> Did something change in the latest version of your library? Should we adjust our runners?

I will take a look

drom:
please try `tree-sitter-verilog` v0.20.0

PR #1382, opened 2021-03-09
Freeze tree-sitter version

tgorochowik:
This does not freeze tree_sitter_verilog, but its dependency.
tree_sitter_verilog seems to be broken with latest tree-sitter, this we
have to freeze its version to make it work again.


tgorochowik:
It fixes the `sv-tests-ci / tree-sitter-verilog` job, merging to unblock other PRs

PR #1399, opened 2021-03-17
WIP: Don't use kokoro

joannabrozek:


tgorochowik:
@hzeller @mithro This is the effort we discussed which is intended to switch the github pages generation to github actions (instead of kokoro).

One thing we need assistance with is generating the tokens and making the available in env so the actions in this repo have privilages to push to the repository with the results. Would you mind helping us set this up?

joannabrozek:
Continued in a separate PR: https://github.com/SymbiFlow/sv-tests/pull/1582

Issue #1400, opened 2021-03-17
Missing define found in verilog config *.vlt in Ibex trips some parsers


hzeller:
Parsers that actually can deal with *.vlt files that is.

There is a *.vlt file in Ibex that expands a `verilator_config` [1] macro that it doesn't see. So either this is something generated from the wrong kind of fusesoc setting, or we need to include the define beforehand.

Example: 
 * sv2v https://symbiflow.github.io/sv-tests-results/#Sv2v_zachjs|ibex|ibex
  * sv-parser https://symbiflow.github.io/sv-tests-results/#sv_parser|ibex|ibex

Interestingly, it works in Verilator: https://symbiflow.github.io/sv-tests-results/#Verilator|ibex|ibex  -- maybe this is just a predefined value there ?

[1]: interesting meta question: how to put a backtick in a back-ticked region in github markdown ?

MikePopoloski:
Surely verilator-specific files should only be passed to verilator and not other tools? I raised a similar issue in #1347

hzeller:
I'll assign both related ones to Tom to have a look.

tgorochowik:
> [1]: interesting meta question: how to put a backtick in a back-ticked region in github markdown ?

I think the easiest way is to use multiple `` ` `` to escape it, for example
``` `` ` `` ``` produces `` ` `` 
```` ``` `` ` `` ``` ```` produces ``` `` ` `` ``` 
etc

MikePopoloski:
I fixed this in #1545

Issue #1347, opened 2021-02-16
Ibex core test passes verilator-specific files to other tools


MikePopoloski:
It looks like the test is passing verilator linting rule files to other tools (such as slang) which results in nonsense errors such as 
```
../../src/github/sv-tests/build/lowrisc_ibex_ibex_core_0.1/src/lowrisc_lint_common_0.1/tools/verilator/common.vlt:7:1: error: unknown macro or compiler directive '`verilator_config'
`verilator_config
^
```

MikePopoloski:
Looks like this was fixed...

PR #1412, opened 2021-03-24
uhdm-yosys: fix testing ibex-core

kamilrakoczy:
This PR fixes building ibex-core using uhdm-yosys.

### DIFF
--- a/generators/fusesoc
+++ b/generators/fusesoc
@@ -79,18 +79,38 @@ for fusesoc_conf_file in fusesoc_conf_files:
     incdirs = ''
     sources = ''
 
-    with open(core_conf_path, "r") as f:
-        for line in f:
-            line = line.strip()
-            if line.startswith("+incdir+"):
-                incdirs += os.path.join(
-                    sources_path,
-                    line.partition("+incdir+")[2]) + ' '
-            elif (core_conf_path.endswith('.vc')
-                  and line.startswith("--top-module")):
-                top_module = line.partition("--top-module ")[2]
-            elif not line.startswith("-") and line:
-                sources += os.path.join(sources_path, line) + ' '
+    if core_conf_path.endswith(".vc") or core_conf_path.endswith(".scr"):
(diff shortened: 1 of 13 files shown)


hzeller:
Since this is refining fusoc to create the file list, would this also fix #1400 and #1347 ?

tgorochowik:
@hzeller since fusesoc generates config meant for specific tools (verilator and yosys in this case) the best thing we can do is adding a filter to make sure the tests only run on those.

That would mean skipping other tools completely. I am not sure if there is a way to generate a generic/tool-agnostic config with fusesoc.
If there is, then this would be a great next step, for now we intend to ignore most of the tools. So you could say those issues will be fixed _in a way_.

generators/fusesoc
```diff
@@ -29,11 +29,26 @@ except IndexError:
...
     sys.exit(1)
 
+try:
+    runner = os.environ['RUNNERS_FILTER']
+except KeyError:
+    runner = None
+
+tests_list_exclude = []
+
+if runner in ["UhdmYosys"]:
```
tgorochowik, line 39: This needs to be moved to the generic runner code (as discussed offline).

The runners need to check whether a specific test is intended for them and then skip it if it's not. This implementation will not work

hzeller:
In this particular case, most tools seem to just be stumbling upon the *.vlt file. Can we hack around that and just filter that out for other tools ?

generators/fusesoc
kamilrakoczy: Done, I added new meta-tag: ``compatible-runners`` to check whether a specific test is for given runner.

kamilrakoczy:
> In this particular case, most tools seem to just be stumbling upon the *.vlt file. Can we hack around that and just filter that out for other tools ?

I've added this hack and now ``*.vlt`` files (for ``ibex`` test) should only go to verilator tools.

tgorochowik (approved):

PR #1426, opened 2021-03-27
Fixes for UVM tests

MikePopoloski:
Fixed a few issues in various UVM tests:
1. strings are not comparable to null, so changed to empty string
2. One test was trying to make a function call statement in the middle of a class body which is invalid
3. The generated uvm_classes_2 tests were not providing parameters to the super class constructor

Signed-off-by: MikePopoloski <mike@popoloski.com>


tgorochowik (approved):
Thanks!

Issue #612, opened 2020-02-03
Add more tests for 11.3.6 Assignment within an expression


tgorochowik:
We need more tests for this chapter, examples:
```
next_id = ++id;
```

or

```
b = (a+=1);
```
Ref: #609

tgorochowik:
Fixed in #1436

PR #1436, opened 2021-04-04
Added tests for 11.3.6

Prahitha:
#612

### DIFF
--- a/tests/chapter-11/11.3.6--assign_in_exp-sim.sv
+++ b/tests/chapter-11/11.3.6--assign_in_exp-sim.sv
@@ -1,0 +1,19 @@
+/*
+:name: assign_in_exp_sim
+:description: assignment in expression simulation test
+:type: simulation parsing
+:tags: 11.3.6
+*/
+module top();
+
+int a;
+int b;
+int c;
+
+initial begin
+        c = a;
+        b = (a-=1);
+        $display(":assert: (%d == %d)", b, (c-1));
+end
+
+endmodule
--- a/tests/chapter-11/11.3.6--assign_in_expression-sim.sv
+++ b/tests/chapter-11/11.3.6--assign_in_expression-sim.sv
@@ -1,0 +1,19 @@
+/*
+:name: assign_in_expression_sim
+:description: assignment in expression simulation test
+:type: simulation parsing
+:tags: 11.3.6
+*/
+module top();
+
+int a;
+int b;
+int c;
+
+initial begin
+        c = a;
+        b = (++a);
+        $display(":assert: (%d == %d)", b, (c+1));
+end
+
+endmodule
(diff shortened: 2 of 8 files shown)


tests/chapter-11/11.3.6--assign_in_exp-sim.sv
```diff
@@ -0,0 +1,19 @@
...
+module top();
+
+int a;
+int b;
+int c;
+
+initial begin
+        c = a;
+        b = (a-=1);
+    $display(":assert: (b == (c-1))");
```
tgorochowik, line 16: Please try to make the indentations the same

tests/chapter-11/11.3.6--assign_in_exp-sim.sv
```diff
@@ -0,0 +1,19 @@
...
+end
+
+endmodule
```
tgorochowik, line 19: Can you add a newline at the end of all the tests?

tgorochowik (changes requested):
Thank you for contribution! Please see the comments below

tests/chapter-11/11.3.6--assign_in_expression-sim.sv
```diff
@@ -0,0 +1,19 @@
...
+module top();
+
+int a;
+int b;
+int c;
+
+initial begin
+        c = a;
+        b = (++a);
+    $display(":assert: (b == (c+1))");
```
tgorochowik, line 16: The log of the simulation is evaluated by python (we look for the `:assert:` statements). To make it work we need to substitute `b` and `c` with their values, see this as an example:
https://github.com/SymbiFlow/sv-tests/blob/740d4ae9ba4e8b40bebc47abc2f2f6e4c5abb391/tests/chapter-7/arrays/unpacked/reduction-methods/sum.sv#L16

Prahitha: So here, it would be `$display(":assert: (%d == %d), b, (c+1))`?

tests/chapter-11/11.3.6--assign_in_exp-sim.sv
Prahitha: I saw a couple of examples and they had the assertion in a separate block, for instance: https://github.com/SymbiFlow/sv-tests/blob/740d4ae9ba4e8b40bebc47abc2f2f6e4c5abb391/tests/chapter-11/11.3.6--assign_in_expr-sim.sv#L14-L17
So I was wondering, is it for syntax or code readability?

tests/chapter-11/11.3.6--assign_in_expression-sim.sv
tgorochowik: yes, that should work

tgorochowik (commented):
These changes look great, thank you for your contribution!

One final request: can you please squash your commits together? Separating changes into separate commits is a good idea if you're adding two distinct things. Here it adds the tests and then a separate commit fixes them, so it would be best to have this as as single commit to make the history cleaner

Prahitha:
Hope I did it right, thanks for the help!

tgorochowik (approved):
LGTM, thanks!

PR #1439, opened 2021-04-09
Don't run PR commenter on master

tgorochowik:
The workflow for the commenter is started not only in pull requests, but also after the PR is merged, e.g: https://github.com/SymbiFlow/sv-tests/actions/runs/729723318

This PR is an attempt to stop that.


tgorochowik:
Let's merge it, I will create a draft PR shortly to see if it didn't influence the code that runs in PRs (as it runs in the context of the main branch it might have..)

Issue #1438, opened 2021-04-09
Add -sverilog to Surelog command line for black-parrot test


alaindargelas:
That will prevent syntax errors for SystemVerilog constructs present in .v files

tgorochowik:
Fixed in #1451

PR #1451, opened 2021-04-15
Add -sverilog to Surelog cmd for black-parrot test

lpawelcz:
This PR is adding -sverilog flag to Surelog for black-parrot tests. Reported in: https://github.com/SymbiFlow/sv-tests/issues/1438

Signed-off-by: Paweł Czarnecki <pczarnecki@antmicro.com>


tgorochowik (approved):
Let's merge it, we used a similar _hack_ here: https://github.com/SymbiFlow/sv-tests/blob/98e67cdfc234179c97cb305600c24fb40bb13b71/tools/runners/UhdmYosys.py#L53

But I believe we should fix both differently, I'll file an issue for that.

PR #1464, opened 2021-04-19
Fix the RSD generator

MikePopoloski:
This fixes the RSD generator to allow tools that actually check correctness of code to elaborate the design.


mithro:
@MikePopoloski -- If you are interested, as a thank you for your contributions, I would be more than happy to send you a small "FPGA development board care package". If you would like this, please email me @ me@mith.ro with your address + phone number and I'll get it shipped to you!

hzeller:
LGTM. Can you fix the formatting issue the [CI worries about](https://github.com/SymbiFlow/sv-tests/pull/1464/checks?check_run_id=2376890424) ? (I think there is a `make format` in the toplevel Makefile to help with that)

PR #1467, opened 2021-04-21
Fixed variable width columns

Prahitha:
Added 45-degree angle and made the headline fixed

https://user-images.githubusercontent.com/44160152/115524316-9f42e200-a2ab-11eb-9020-6915bf686257.mp4

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -48,7 +48,7 @@
     <p class="headline"><a href="https://github.com/SymbiFlow/sv-tests"><em>Test suite to check compliance with the SystemVerilog LRM by chapter as well as some real-world cores and test-cases.</em></a></p>
     <table id="report_table">
       <thead>
-        <tr>
+        <tr class="fixed_header">
           <th>  </th>
           <th>  </th>
           {% for tool, tooldata in report|dictsort %}
--- a/conf/report/report.css
+++ b/conf/report/report.css
@@ -11,6 +11,22 @@ body {
     text-decoration: none;
 }
 
+.fixed_header {
+  height: 150px;
+}
+
+table.dataTable thead .sorting, table.dataTable thead .sorting_asc, 
+table.dataTable thead .sorting_desc, table.dataTable thead .sorting_asc_disabled, 
+table.dataTable thead .sorting_desc_disabled {
+  background-position: bottom right;
+  max-width: 50px;
+}
+
+.tool_link {
+  display: block;
+  transform: translate(0px, 30px) rotate(315deg);
+}
+
 table.dataTable {
   width: auto;
   table-layout: fixed;


conf/report/report.css
```diff
@@ -3,17 +3,36 @@ body {
 }
 
 .headline {
-    text-align: center;
+  width: 100%;
+  text-align: center;
+  position: fixed;
+  display: table-row;
```
tgorochowik, line 9: This is what causes the header to overlap with the table, please remove this.

conf/report/report.css
```diff
@@ -3,17 +3,36 @@ body {
...
 .headline a {  /* Only subtly hint that this is a link */
-    color: #00006f;
-    text-decoration: none;
+  color: #00006f;
+  text-decoration: none;
+}
+
+label {
+  position: fixed;
+  right: 0;
```
tgorochowik, line 19: This causes the search input to be overlapped with the table, we don't want the position to be fixed, we want to be able to scroll down and leave it behind at this point

conf/report/report-template.html
```diff
@@ -44,7 +44,7 @@
   </style>
 
   <body>
-    <h1 class="headline"><a href="https://github.com/SymbiFlow/sv-tests">SV-Tests</a></h1>
+    <h1 class="headline"><a href="https://github.com/SymbiFlow/sv-tests">SV-Tests</a></h1><br/><br/><br/>
```
tgorochowik, line 47: Why??

conf/report/report.css
```diff
@@ -3,17 +3,36 @@ body {
...
 }
 
 table.dataTable {
   width: auto;
   table-layout: fixed;
+  padding-top: 120px;
```
tgorochowik, line 25: I assume this was to make the text in the headers stop overlapping with the titles at the top - this is not a proper solution, this is what causes everything to flicker when starting to scroll down.

tgorochowik (changes requested):
Thanks for the PR! Looks promising, although there are some issues with it, please see the comments below.

tgorochowik:
In addition:
* I Think that to properly solve this, you need to force the rows in the header to have a height that matches the text inside them (the padding you added made it look like it was okay but the text is still outside the row).
* Please keep the original formatting of the sources (don't change the indentation)

conf/report/report.css
Prahitha: I understand, however, when the headline is fixed and the search box is moving, they were overlapping on the x-axis. Maybe it was because I didn't use the right approach to fix the headline. I'll look into this, thanks!

Prahitha:
Thanks for the feedback! I inspected the table header in the browser and its height didn't change. I'll try to review more carefully before submitting a PR next time (sorry for changing the indentation! :/)

conf/report/report.css
tgorochowik: Oh, okay, I understand what you wanted to do here then - unfortunately it causes the issues we can see when scrolling down.

It's nice that you tried to fix it, but please keep one feature per PR - otherwise it is hard to tell which code change is expected to achieve what.

And as for this change, maybe we could move the horizontal scroll to be inside the table (as opposed to scrolling the whole page like it happens now): https://datatables.net/examples/basic_init/scroll_x.html

(but let's do this separately, not as a part of this PR)

Prahitha: Sure! I'll work on only fixing the variable width columns for now then. I've also just realised that the column width still changes on sorting, so I'll look into it and push the changes 😄

tgorochowik (commented):
I see the height is now hard-coded to a larger value - it is not perfect but I think it should be acceptable - it looks good!

Could you please squash your commits and mark your PR as ready?

tgorochowik (approved):
LGTH, thanks!

PR #1471, opened 2021-04-23
Allow build cores without `-noelab` argument for Surelog tool

joannabrozek:
Add the possibility to build cores without  `-noelab` argument for Surelog tool.


tgorochowik (approved):
LGTM!

PR #1482, opened 2021-04-25
Blacklist broken yosys tests

MikePopoloski:
Blacklisting tests because:
* The hana "test_parser" test has messed up syntax around a macro expansion; it seems like the test is assuming the tokens will automatically concatenate with each other. No commercial tool I tried can parse this and the LRM does not seem to allow it either.
* memories_issue00335 references a parameter before it's declared
* simple_func_width_scope uses a function declared inside a generate block in a constant expression. This is directly disallowed by the LRM.
* simple_named_genblk references an unnamed generate block by name which is not allowed (and makes no sense anyway).
* sva_basic04 tries to bind a module to a "top" instance even though there is no such instance.
* sva_basic05 tries to instantiate a module called "demo" even though there is no such module.
* svinterface1 and svinterface_at_top both pass the wrong modport to an interface port, which is disallowed.

### DIFF
--- a/conf/generators/meta-path/yosys-memories.json
+++ b/conf/generators/meta-path/yosys-memories.json
@@ -4,5 +4,6 @@
 	"paths": [
 		["tools", "yosys", "tests", "memories"]
 	],
-	"matches": ["*.v"]
+	"matches": ["*.v"],
+	"blacklist": ["issue00335.v"]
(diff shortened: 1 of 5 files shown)


tgorochowik (approved):
Thank you for verifying these!

PR #1483, opened 2021-04-25
Fix basejump STL test setup

MikePopoloski:
* Add various missing package references
* Add missing include directory
* Blacklist the package files, since otherwise the package tests will include the packages multiple times causing errors


MikePopoloski:
Bump on this? I don't think this is in any way controversial, someone just needs to merge it.

mithro:
@kgugala / @hzeller - Can you make sure someone takes a look at this? Thanks!

tgorochowik (commented):
LGTM in general, but it produces new failures in yosys/yosys-sv which are not seen anywhere else (the remaining numbers in the table are consistent with other PRs - the table cannot be fully trusted yet).

I'd like to take a look at those yosys tests to see wha'ts going on, that's why it's not merged yet

Issue #1470, opened 2021-04-21
add horizontal scroll for the dataTable


Prahitha:
Having the table as scrollable rather than the whole page/report

PR #1484, opened 2021-04-26
fixed first column of the table

### LINKED ISSUE #1469: Keep the first column (tags) fixed on scroll-x
Similar to the fixed header, having the first column fixed would make it easier to browse the logs/report

Prahitha:
fixes #1469 and #1470

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -107,28 +107,28 @@
           {% endfor %}
         </tr>
         <tr class="report_summary_row">
-          <th class="report_table_info" colspan="2"> User time elapsed </th>
+          <th class="report_table_info gray_header" colspan="2"> User time elapsed </th>
           {% for tool, tooldata in report|dictsort %}
           <th class="report_table_result" title="{{ tool.lower() }}" >
             {{'%0.0f'| format(report[tool]["user_time"])}}s </th>
           {% endfor %}
         </tr>
         <tr class="report_summary_row">
-          <th class="report_table_info" colspan="2"> System time elapsed </th>
+          <th class="report_table_info gray_header" colspan="2"> System time elapsed </th>
           {% for tool, tooldata in report|dictsort %}
           <th class="report_table_result" title="{{ tool.lower() }}" >
             {{'%0.0f'| format(report[tool]["system_time"])}}s </th>
           {% endfor %}
         </tr>
         <tr class="report_summary_row">
-          <th class="report_table_info" colspan="2"> Maximum ram usage </th>
+          <th class="report_table_info gray_header" colspan="2"> Maximum ram usage </th>
           {% for tool, tooldata in report|dictsort %}
           <th class="report_table_result" title="{{ tool.lower() }}" >
             {{'%0.0f'| format(report[tool]["ram_usage"])}} MB </th>
           {% endfor %}
         </tr>
 	<tr class="report_summary_row">
-          <th class="report_table_info" colspan="2"> Average throughput passed for inputs &gt; 1KiB </th>
+          <th class="report_table_info gray_header" colspan="2"> Average throughput passed for inputs &gt; 1KiB </th>
           {% for tool, tooldata in report|dictsort %}
           <th class="report_table_result" title="{{ tool.lower() }}" >
             {{'%0.0f'| format(report[tool]["passed_throughput"]|float)}} KiB/s </th>
--- a/conf/report/report.css
+++ b/conf/report/report.css
@@ -15,6 +15,36 @@ body {
   height: 150px;
 }
 
+.report_table_info {
+  position: sticky;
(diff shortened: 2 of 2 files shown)


conf/report/report-template.html
```diff
@@ -23,7 +23,10 @@
     $(document).ready( function () {
       $('#report_table').DataTable( {
         paging: false,
-        fixedHeader: true,
+        scrollX: true,
+        fixedColumns: {
+          leftColumns: 1,
```
tgorochowik, line 28: Can we freeze two columns? It looks much better then

tgorochowik (commented):
Look good in general! Thank you for the PR!

One thing that feels weird is that the horizontal scroll bar is not visible unless we scroll down to the bottom of the page.

Could you try to make it stick to the bottom of the page in a similar way the header sticks to the top of the page?

Prahitha:
When using fixedColumns, we have to add the `scrollX: true` which now makes only the table scrollable instead of the whole page (please correct me if I am wrong here). 

Since only the table is scrollable (not the whole page), the scrollbar is attached with the table only (i.e. I can only place the scrollbar to appear after the header, body, or footer), which is probably why the position seems weird now. Please let me know what to think. Thanks!

mglb:
> When using fixedColumns, we have to add the `scrollX: true` which now makes only the table scrollable instead of the whole page (please correct me if I am wrong here).
> 
> Since only the table is scrollable (not the whole page), the scrollbar is attached with the table only (i.e. I can only place the scrollbar to appear after the header, body, or footer), which is probably why the position seems weird now. Please let me know what to think. Thanks!

I think it can be solved with pure CSS (i.e. without using `scrollX: true` and `fixedColumns` script) with something like:
```
position: sticky;
left: 0 /* or some appriopriate value for second column + margins, etc */
```
applied to cells of first and second column (`.report_table_info` and `.report_table_tag`).

Overall:
It would be nice to enable fixed columns only when the browser width is larger than those two columns.

mglb:
@Prahitha setting header's z-index to value larger than fixed column's z-index should fix the problem visible in the video.

Prahitha:
Yes, that's how it is right now, I should've pushed my code I guess. So, the z-index for the columns is 1 and for thead it is 2. I've also tried with higher values like 2 and 4 for the columns and the header respectively.

mglb:
@Prahitha Hint from Firefox:
Screenshot_20210513_121032

However, the floating header is placed in its own, separate table element (the one with class `fixedHeader-floating`). Setting `z-index: 4` on it should work.

Prahitha:
Ah, thanks for the tip, it works now! I didn't see any `fixedHeader-floating` class while inspecting in the browser but now that you mentioned it, I see it in the source code for css of the fixedHeader extension. Is there some reason for this or am I missing something? Please let me know, thank you!

mglb:
This extra table with `fixedHeader-floating` class and floating header inside is created by the script when the page is scrolled down (i.e. when the floating header is activated), and removed when it is no longer needed (i.e. page scrolled to the top).

Looks good so far, except that the `Total *` cells on the bottom are not fixed:
Screenshot_20210513_182618-1

Prahitha:
Understood, thanks for the clarification!

Prahitha:
So, I just made the fixes. I also observed that the color of the cells with fixed columns in the last 4 rows, in my previous commit, didn't match the colors on the original website. So I added a class called `gray_header` for those, hope that it's fine. Please let me know, thanks!

(adding images for reference)
my previous commit
on the original website

conf/report/report.css
```diff
@@ -15,6 +15,36 @@ body {
...
+.report_table_info {
+  position: sticky;
+  left: 0;
+  z-index: 2;
+  background-color: white;
+}
+
+.report_table_tag {
+  position: sticky;
+  left: 31em;
```
mglb, line 27: `calc()` should be used here to sum first column's width and padding. Otherwise there is slight misalignment:
Screenshot_20210514_153113

mglb (commented):
A few small adjustments and it will be ready:
* misalignment
* is there some easy way to restore the border between white and gray headers?

Prahitha:
I'm not sure if this is the best approach, please let me know what you think. And thanks for your help and being patient throughout!

conf/report/report.css
```diff
@@ -71,6 +71,10 @@ table.dataTable tfoot th {
   outline: 2px solid #505050;
 }
 
+.first_row {
```
mglb, line 74: I would use:
```css
.report_footer > tr:first-of-type > .report_summary_header
```
In order to explicitly tell the reader which element it applies to.

mglb:
@Prahitha good job!

@tgorochowik LGTM.

tgorochowik (approved):

PR #1545, opened 2021-05-24
Fix ibex test

MikePopoloski:
This makes it possible for simulation tools to build the Ibex core (and changes slang to not use the synthesis target).
- vlt and cc files need to be filtered out for all non-verilator tools
- Even Verilator couldn't build the Ibex core because the fusesoc generator wasn't including the necessary defines

### DIFF
--- a/tools/runner
+++ b/tools/runner
@@ -209,6 +209,11 @@ for key in libs.keys():
 
 test_params['defines'] = test_params['defines'].split()
 
+# Some tests have *.vlt and *.cc files which are specific to verilator,
+# so filter them out for all other runners.
+if 'verilator' not in args.runner:
+    test_params['files'] = [f for f in test_params['files'] if not f.endswith('.vlt') and not f.endswith('.cc')] 
+
 try:
     tmp_dir = tempfile.mkdtemp()
 except (PermissionError, FileExistsError) as e:
--- a/tools/runner
+++ b/tools/runner
@@ -209,13 +209,11 @@ for key in libs.keys():
 
 test_params['defines'] = test_params['defines'].split()
 
-# Some tests have *.vlt and *.cc files which are specific to verilator,
-# so filter them out for all other runners.
-if 'verilator' not in args.runner:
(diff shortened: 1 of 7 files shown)


tools/runner
```diff
@@ -209,6 +209,14 @@ for key in libs.keys():
...
 test_params['defines'] = test_params['defines'].split()
 
+# Some tests have *.vlt and *.cc files which are specific to verilator,
+# so filter them out for all other runners.
+if 'verilator' not in args.runner:
+    test_params['files'] = [
+        f for f in test_params['files']
+        if not f.endswith('.vlt') and not f.endswith('.cc')
+    ]
+
```
tgorochowik, line 219: Hmm, we have indeed allowed similar hacks in some of the wrappers for specific runners, but this is a common file for all runners and I'd rather keep it cleaner (i.e. it should not be aware of any particular runners).

Maybe instead of this, we could create an "allow list" of file extensions. The base runner could then define the default allow list as `[".sv", ".v", ".vh", ".svh"]` (and others if relevant).

The verilator runners could then extend the list with `.vlt` and `.cc`.

And here, instead of your original change, we could filter out the files list based on the allow list.

MikePopoloski: Sure, I can look at that.

tools/runner
```diff
@@ -209,6 +209,12 @@ for key in libs.keys():
 
 test_params['defines'] = test_params['defines'].split()
 
+# Filter test files based on what the runner claims to support.
+test_params['files'] = [
+    f for f in test_params['files']
+    if os.path.splitext(f)[1] in runner_obj.allowed_extensions
+]
+
```
tgorochowik, line 217: Could we change it to a regular for so that we can show a `self.info` message or maybe a warning when skipping a file? It works fine now but may make our lives easier if something is unexpectedly skipped in the future

tgorochowik (approved):
LGTM, thanks for the changes, one last minor request

tgorochowik (approved):
Thanks!

PR #1547, opened 2021-05-24
Fix clocking block test, add a should_fail test to mirror

MikePopoloski:
One of the clocking block tests was invalid because it tried to assign to nets from within a procedural context. This diff adds a should_fail to that test and adds a new duplicated test with the original error fixed.


tgorochowik (approved):
Thanks!

PR #1549, opened 2021-05-25
filter for test coverage

Prahitha:
note: the filters will temporarily be at the end of the table -> `$(column.footer()).empty()`
it is not fully implemented yet. the column.draw() inside the conditional statements doesn't work and needs some work.

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -15,6 +15,8 @@
     <script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.10.19/js/jquery.dataTables.min.js"></script>
     <script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/fixedheader/3.1.5/js/dataTables.fixedHeader.min.js"> </script>
     <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/fixedheader/3.1.5/css/fixedHeader.dataTables.min.css">
+    <script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/select/1.3.3/js/dataTables.select.min.js"></script>
+    <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/select/1.3.3/css/select.dataTables.min.css">
     <link rel="stylesheet" type="text/css" href="report.css">
     <script type="text/javascript" charset="utf8" src="report.js"></script>
   </head>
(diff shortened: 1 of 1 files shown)


Prahitha:
Note: Still a WIP
Here is a small demo of what it does, there are a few bugs (like the one I paused at in the middle of the video)
https://user-images.githubusercontent.com/44160152/120211208-4cffc400-c24e-11eb-8011-6da912c8f8da.mp4

(the select dropdown was not recorded in the screen recording)
For the first column, I selected all (i.e., all test cases passed) and then selected 0 for the next column. This basically filters from the result of the previous one, if not cleared. To clear an active filter, the empty select option has to be selected.

conf/report/report-template.html
```diff
@@ -33,7 +35,50 @@
           {% for tool in report %}
           { "orderDataType": "test-status" },
           {% endfor %}
-        ]
+        ],
+          initComplete: function() {
+            this.api().columns([2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17]).every(function() {
```
tgorochowik, line 40: what exactly does this do? do we have to have that list literal here?

tgorochowik (commented):
LGTM in general, nice work!

I am still not sure about putting this in the cell it is now, it seems to limit the visibility of the total number of tests passed?

also: what happens if you select something for multiple cells? is this a logical and? or a logical or? it is not immediately clear I think

tgorochowik:
CC @hzeller

conf/report/report-template.html
Prahitha: so I listed them out to not have the dropdown for the tags (first 2 columns) but I'll see if this can be done in some other way

Prahitha:
thanks! so the position of the drop-downs definitely has to change and I planned on working on it once the functionality seemed okay. and, I believe it's a logical and because it filters out from the previous result only. i'll look into other ways if it's not as intuitive, please let me know

conf/report/report-template.html
hzeller: So this will have to be changed whenever we add a new tool ? This sounds brittle. Can this be formulated as 'everything but the first two columns' instead ?

conf/report/report-template.html
```diff
@@ -33,7 +35,50 @@
...
+                var column = this;
+                var select = $('<select><option value="Select"></option></select>')
```
hzeller, line 42: General question to consider: instead of having in-line implementation of functions, putting them in a separate JS file might be easier to maintain in the long term.
It is easy to accumulate important code that is hard to understand if buried in a large HTML template (not sure what are good practices are to break this up, as it seems that the closure takes a few values from the context (such as `$(this).val()`) - so possibly the function to be implemented externally needs explicit parameters; but that is good for documentation).

So while this looks not a large problem here at this point (and maybe the conclusion might be that this particular use might be benign), it might become some problem in the future. So worthwhile considering a general plan for code structuring.

conf/report/report-template.html
Prahitha: I believe this is resolved in the latest commit. I'll also look into adding a separate JS file for the script. Thanks for the suggestions!

Prahitha:
below is a demo of how the toggle looks and works. also, just to clarify, the filtering is currently logical and but can support logical or too. any suggestions on how to implement it differently are welcome. thanks!

https://user-images.githubusercontent.com/44160152/120515413-e9130200-c3eb-11eb-9729-75fc00afabd5.mp4

tgorochowik (commented):
Adding it this way will not allow us to easily mix the filter options with other filter types (once we add them).

"filter type" should be a separate field, so once we click "show filter options" we should get a combo box with "filter type" (for now we should be able to only select "coverage" but other filter types should be possible to select in the future.

Then once you do select "coverage" another two inputs should appear - one for the tool selection and another one for coverage selection.

below all of that we should have two buttons - "apply" and "add"

"apply" is self-explanatory (and can be optional - it is okay to automatically apply everything once the row i actually correct - I think this is done like this at the moment)
"add" should generate another row of fields which we can use to combine the current filter (the first row) with another filter (test coverage for another tool for now, other filter types in the future).

Is that explanation clear? Or do you need more details?

Another option - a possibly more modern approach - would be to make the search bar recognize some keywords and make it apply the filters based on what is typed into the search bar - something similar to what github and gitlab do: https://docs.gitlab.com/ee/user/search/ (for a start we could use the google-like `filter_type:filter_input` notation) - but I think this may be harder to implement

@hzeller what do you think?

Prahitha:
Thanks for the detailed explanation! I've attached an image below to confirm if I'm understanding properly. Please let me know, thanks
image

tgorochowik:
Hmm, not exactly, please see the attached diagram, the steps are split to separate drawings, hopefully it will be easier to understand what I mean
sv tests filtering

PR #1554, opened 2021-05-28
minor ui fix

Prahitha:
When using search, the width/position of the first column is not consistent. This PR fixes the issue by making the width of the first column 30em always.

Previously (when the first column is not long/wide enough):
Screenshot (33)


tgorochowik:
Could you please add a screenshot or at least a description of what the fix actually does?

tgorochowik (approved):
Indeed, thanks for the fix!

PR #1565, opened 2021-05-31
Fix Ariane core build

MikePopoloski:
This fixes the Ariane build for the slang runner. Two problems fixed:
1. The Ariane target we're using doesn't build if you don't define VERILATOR, since it otherwise references nonexistent modules.
2. The filelist includes "axi2apb_wrap.sv" which tries to use nonexistent interfaces. axi2apb_wrap is never used anywhere itself so it's fine to just remove it.

### DIFF
--- a/generators/ariane
+++ b/generators/ariane
@@ -69,7 +69,12 @@ for components in ver_cmd_comp:
         top_module = args
         get_top_module = False
     elif args.endswith('.sv'):
-        sources += os.path.join(ariane_path, args) + ' '
+        # The Ariane file lists contains a reference to this file but the module itself
+        # isn't instantiated anywhere. Problematically, it contains references to a
+        # nonexistent interface, so it otherwise breaks elaboration. This hack can be
+        # removed if the problem gets fixed upstream.
+        if 'axi2apb_wrap.sv' not in args:
+            sources += os.path.join(ariane_path, args) + ' '
 
 os.chdir(main_path)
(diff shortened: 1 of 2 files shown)


tgorochowik (approved):

generators/ariane
```diff
@@ -69,7 +69,12 @@ for components in ver_cmd_comp:
         top_module = args
         get_top_module = False
     elif args.endswith('.sv'):
-        sources += os.path.join(ariane_path, args) + ' '
+        # The Ariane file lists contains a reference to this file but the module itself
+        # isn't instantiated anywhere. Problematically, it contains references to a
+        # nonexistent interface, so it otherwise breaks elaboration. This hack can be
+        # removed if the problem gets fixed upstream.
+        if 'axi2apb_wrap.sv' not in args:
+            sources += os.path.join(ariane_path, args) + ' '
```
tgorochowik, line 77: I think we use the makefile provided by Ariane to generate this project here. Is this a problem that is maybe worth reporting in mainline Ariane? or is this specific to the runner wrapper here?

MikePopoloski: Yeah, I suspect this is a problem upstream. I can file an issue with them to see what they think but in the meantime I think we should land this here.

tgorochowik: sure, thanks!

MikePopoloski: https://github.com/openhwgroup/cva6/issues/672

PR #1580, opened 2021-06-08
Remove bp_lce from black-parrot test list

MikePopoloski:
The bp_lce module is not valid as a top-level target; it has parameters that are defaulted to "inv" which results in invalid accesses later in the code. The module needs to be instantiated as part of larger design with valid parameter values, so it doesn't make sense to test in isolation.

### DIFF
--- a/generators/black-parrot
+++ b/generators/black-parrot
@@ -68,10 +68,6 @@ lists = [
         'top': 'bp_cce',
         'flist': 'bp_top/syn/flist.vcs'
     }, {
-        'name': 'bp_lce',
-        'top': 'bp_lce',
-        'flist': 'bp_top/syn/flist.vcs'
-    }, {
         'name': 'bp_uce',
         'top': 'bp_uce',
         'flist': 'bp_top/syn/flist.vcs'


taylor-bsg:
I think rather than removing this, we can just fix in the bp repo. We have a methodology that was hashed out by other sv-tests participants that is a more correct way of signaling an abstract module. after our tapeout @dpetrisko can implement it..

MikePopoloski:
Isn't this module already included in bp_multicore and bp_unicore (and maybe the bp_be and bp_fe top-level targets) as well? I'm not sure why it needs to included in yet another test case as a top-level, especially since it doesn't actually work in this configuration anyway.

MikePopoloski:
I have no problem if you want to fix this upstream and re-add it later, but I think this should be removed in the meantime since it is broken and it seems like the time for a fix to make it to master in black-parrot is long (not a criticism, obviously you need to be very careful for tapeout).

MikePopoloski:
This is still busted by the way...

taylor-bsg:
This is fixed in the dev branch which will be merged to master soon…

@dpetrisko

M

dpetrisko:
Bump to master requires validation with our FPGA and ASIC flows, probably a few weeks out (sept 1st to put a self imposed deadline). Could also just point the submodule to current dev?

- Dan

MikePopoloski:
Pointing the submodule at dev sounds like a good idea in general anyway. Perhaps I'll do that if no one has any objections?

tgorochowik:
+1

PR #1582, opened 2021-06-10
Use GH Actions to update reports

joannabrozek:
Use GH Actions to update reports in the sv-tests-results repository.


tgorochowik:
Closing as CI doesn't seem to start here.

PR #1583, opened 2021-06-10
ui for filtering; no functionality yet

Prahitha:
Note: This is a WIP and does not have the filtering functionality yet (the applyFilter function has to be implemented). Ideally, the test types should also not be hard-coded (will work on it). 

Any suggestions on how to store the selected options properly would be really helpful. Since there are a lot of things to consider, like:
- selecting the percentage first and then the tool vs selecting the tool first and then the percentage (also, how to store these percentages for each tool? would it be a dict?)
- updating the values appropriately if someone goes back to change the values in a previous dropdown (would changing the values at that index be a good approach?) etc
Thanks!

https://user-images.githubusercontent.com/44160152/121487940-372d9400-c9f0-11eb-84bd-a70ad57e7c6c.mp4

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -65,13 +65,80 @@
       $('#filter_wrapper').remove();
       iter = 1;
       var div = $('<div id="filter_wrapper"><button id="mainButton" onclick="addOptions()">Add</button></div>').insertAfter('#headline');
-      filter.length = 0;
+      filters.length = 0;
     }
 
     function applyFilter() {
       apply_active = true;
-      if(!document.getElementById('removeButton'));
+      if(!document.getElementById('removeButton'))
         var remove_all = $('<button id="removeButton" onclick="removeAll()">Remove all filters</button>').insertAfter('#applyButton');
+      
+      report = $('#report_table').DataTable();
+      var num_filters = filters.length;
+      var coverage, coverage_operator, type, type_operator, tool, tool_operator;
+      for(var i=0; i<num_filters; i++){
+        if(filters[i].mode === 'coverage'){
+          coverage_operator = filters[i].condition;
+          coverage = filters[i].value;
+        }
+        if(filters[i].mode === 'type'){
+          type_operator = filters[i].condition;
+          type = filters[i].value;
+        }
+        if(filters[i].mode === 'tool'){
+          tool_operator = filters[i].condition;
+          tool = filters[i].value;
+        }
+      }
+
+      report.columns().every(function() {
+        var column = this;
+        if(column.index() !== 0 && column.index() !== 1) {
+          var theadname = column.header().textContent.trim();
+
+          if(tool_operator === 'is'){
+            if(theadname === tool){
+              var array = [];
+              coverage = parseInt(coverage);
+              column.data().each(function(d, j) {
+                output = d.split('/');
+                tests_passed = output[0];
+                total_tests = output[1];
+                percentage = parseInt((tests_passed/total_tests)*100);
+                
+                console.log(coverage_operator, coverage)
+                if (coverage_operator === '<'){
+                  if(percentage < coverage){
+                    array.push('^'+d+'$');
+                  }
+                }
+                if (coverage_operator === '<='){
+                  if(percentage <= coverage){
+                    array.push('^'+d+'$');
+                  }
+                }
+                if (coverage_operator === '>'){
+                  if(percentage > coverage){
+                    array.push('^'+d+'$');
+                  }
+                }
+                if (coverage_operator === '>='){
+                  if(percentage >= coverage){
+                    array.push('^'+d+'$');
+                  }
+                }
+                if (coverage_operator === '='){
+                  if(percentage === coverage){
+                    array.push('^'+d+'$');
+                  }
+                }
+              });
+              array = array.join('|');
+              column.search(array, true, false, true).draw();
+            }
+          }
+        }
+      });
     }
     
     function addOptions() {
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -15,13 +15,67 @@
     <script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.10.19/js/jquery.dataTables.min.js"></script>
     <script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/fixedheader/3.1.5/js/dataTables.fixedHeader.min.js"> </script>
     <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/fixedheader/3.1.5/css/fixedHeader.dataTables.min.css">
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.3/css/all.min.css" rel="stylesheet"/>
     <link rel="stylesheet" type="text/css" href="report.css">
     <script type="text/javascript" charset="utf8" src="report.js"></script>
   </head>
 
   <script>
+    function Filter(id, mode, condition, value) {
+      this.id = id;
+      this.mode = mode;
+      this.condition = condition;
+      this.value = value;
+    }
+
     var iter = 1;
+    var filters = [];
+    var value;
+    var apply_active = false;
+
+    function removeFilter(object){
+      var current_value = object.id;
+      current_value = parseInt(current_value.charAt(current_value.length-1));
+      var index = filters.findIndex(obj => obj.id === current_value);
+      if(current_value === iter-1 || current_value === iter-2){
+        $('#new_filter_options_'+current_value).remove();
+        $('#remove_'+current_value).remove();
+        current_object = filters[index];
+        mode = current_object.mode;
+
+        if(mode === "coverage"){
+          $('#operator_filter_'+current_value).remove();
+          $('#coverage_filter_'+current_value).remove();
+        }
+        if(mode === "type"){
+          $('#type_condition_filter_'+current_value).remove();
+          $('#type_filter_'+current_value).remove();
+        }
+        if(mode === "tool"){
+          $('#tool_condition_filter_'+current_value).remove();
+          $('#tool_filter_'+current_value).remove();
+        }
+      }
+      else
+        $('#new_options_wrapper_'+current_value).remove();
+      filters.splice(index, 1);
+    }
+
+    function removeAll() {
+      $('#filter_wrapper').remove();
+      iter = 1;
+      var div = $('<div id="filter_wrapper"><button id="mainButton" onclick="addOptions()">Add</button></div>').insertAfter('#headline');
+      filter.length = 0;
+    }
+
+    function applyFilter() {
+      apply_active = true;
+      if(!document.getElementById('removeButton'));
+        var remove_all = $('<button id="removeButton" onclick="removeAll()">Remove all filters</button>').insertAfter('#applyButton');
+    }
+    
     function addOptions() {
+      var tool, operator, percentage, type;
       var toolnames = [];
       $("#report_table thead tr th a").each(function(){
           toolnames.push(this.innerHTML);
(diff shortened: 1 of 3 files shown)


mglb:
> selecting the percentage first and then the tool vs selecting the tool first and then the percentage

Can you make them independent? This way it would be possible to:

* Show results with specified coverage (`coverage ("<" | ">=") (0..100)`)
* Show results for specified tool (`tool ("is" | "is not") TOOL`)
* Show results where specified tool has specified coverage (`coverage ("<" | ">=") (0..100)` and `tool ("is" | "is not") TOOL`)

Such filter could be stored as an array where each item is a tuple/object with 3 fields: `variable` `operator` `value` (e.g. `coverage`, `<`, `40`).

> updating the values appropriately if someone goes back to change the values in a previous dropdown (would changing the values at that index be a good approach?)

Don't automatically change user choices. Highlighting values that became invalid instead. E.g.

* User sets "test type" to parsing
* User sets "tool" to (some tool that supports only parsing)
* User changes "test type" to preprocessing -> drop down with invalid tool name is highlighted in red; text "invalid value" appears next to it.

Features worth considering:
* Ability to specify operator between values of the filter. This would make it possible to e.g. see tools that support parsing AND preprocessing, or tools that support either parsing OR preprocessing. Visually this can look like: `Test type: [Parsing ↓] [or ↓] [      ↓] ...`
* Textual representation of the filter, like `(type=="parsing" or type=="preprocessing") and tool=="surelog" and coverage<="50%"`) - I guess there are javascript libs for parsing such expressions.

Prahitha:
Thanks for the response and suggestions! However, I can't seem to visualize how it should look like. If time permits, could you please make a lo-fi wireframe if possible. Thank you

tgorochowik:
I agree with Mariusz - this way it will be easier to implement and possibly enable us to be even more flexible with the filtering.

One potential issue is the way we display the results, please see the updated graph:

sv tests filtering

Is this clear now?

This way for each applied filter you will ALWAYS have exactly the same set of parameters to store:
* `mode`
* `condition`
* `value`

`condition` and `value` will change depending on `mode` and will have to be handled accordingly, but at least you will have a constant number of arguments which should be easy to store.

Prahitha:
Thanks, I don't completely understand what happens after the 2 boxes for coverage are filled though. Should the next filter appear automatically (here, for mode)? What if the user only wants to filter for coverage?

tgorochowik:
I've added some colors to the graph. The yellow boxes have to be filled before the "apply" button can be clicked.

If user only selects coverage, all tools should be filtered and the coverage should be verified for each of the tools.


One thing to figure out is how to highlight the results.

I think that it is okay to completely hide rows that don't have  any matches (like the search does now). But if we get rows in which not all of the columns have a match, we should find a way to highlight the matches - maybe that could be done with a border?
At the moment we have a large blue border for the selected cell, maybe we could add a yellow/orange border for the cells with matches?

mglb:
The "add" button is redundant - we can always show one empty filter row at the bottom. When user fills it, a new empty row will appear.


* Initial state:
  ```
  [            ▾]
  ```
  (effective filter: "show everything")


* User selects "Coverage". The boxes that appear should have such default values so that the effective filter does not change (e.g. "Coverage >= 0", "Mode is (empty)").
  ```
  [ Coverage   ▾] [ >=  ▾] [ 0  ▾]%
  [            ▾]
  ```
  (effective filter: `coverage >= 0%`, so basically "show everything")


* User changes percentage:
  ```
  [ Coverage   ▾] [ >=  ▾] [ 30  ▾]%
  [            ▾]
  ```
  (effective filter: `coverage >= 30%`)


* User selects "Mode" on the second row.

  ```
  [ Coverage   ▾] [ >=  ▾] [ 30  ▾]%
  [ Mode       ▾] [ is  ▾] [                ▾]
  [            ▾]
  ```
  (effective filter: still `coverage >= 30%`)

* User picks "preprocessing":
  ```
  [ Coverage   ▾] [ >=  ▾] [ 30  ▾]%
  [ Mode       ▾] [ is  ▾] [ Preprocessing  ▾]
  [            ▾]
  ```
  (effective filter: `coverage >= 30% AND mode == preprocessing`)

Prahitha:
Note: Still a WIP
Below, I've attached a video of how it currently works/looks and an image of how the values are stored (before and after clicking the remove icon). Any feedback is welcome, thanks!

https://user-images.githubusercontent.com/44160152/122386360-70837800-cf8b-11eb-97f9-3275d755222b.mp4
Before clicking remove
After clicking remove

Prahitha:
This commit works for very basic filtering - i.e., for toolname "is" {some tool} and coverage only. 
Issues to work on -
- When a filter is active, if we change the value of any dropdown and click on apply, it currently filters from the previous output (i.e, from the dataTable that we get after the first filter is applied)
I've tried adding this:
`report = $('#report_table').DataTable();
 report.search('').draw();`
at the beginning of applyFilters() function to clear any current filter but it didn't seem to make any change
- Filter by classname for type (the test types are added as classes to each cell)

conf/report/report-template.html
```diff
@@ -65,13 +65,80 @@
...
+      filters.length = 0;
     }
 
     function applyFilter() {
       apply_active = true;
-      if(!document.getElementById('removeButton'));
+      if(!document.getElementById('removeButton'))
         var remove_all = $('<button id="removeButton" onclick="removeAll()">Remove all filters</button>').insertAfter('#applyButton');
+      
+      report = $('#report_table').DataTable();
```
Prahitha, line 76: When I tried to declare this as a global variable, i.e., outside of all the functions - it was showing null

conf/report/report-template.html
```diff
@@ -65,13 +65,80 @@
...
+      }
+
+      report.columns().every(function() {
+        var column = this;
+        if(column.index() !== 0 && column.index() !== 1) {
+          var theadname = column.header().textContent.trim();
+
+          if(tool_operator === 'is'){
+            if(theadname === tool){
+              var array = [];
```
Prahitha, line 101: Currently, the filter only works for "is" condition because the array is declared in inside the `report.columns().every()` and will cannot store the values from the previous columns. When I initialized the array outside the function, it did not have access to it and gave an error at `array.push()`

conf/report/report-template.html
mglb: That's because your code runs before the table is created. The table is created below in `$(document).ready( function () {...}`.

conf/report/report-template.html
mglb: You probably had access to the array initialized outside, but you're replacing it with a string in line 136 (`array = array.join('|');`)

conf/report/report-template.html
mglb: That's because your code runs before the table is created. This happens in a callback created in line 279 `$(document).ready(function () {...})`.

conf/report/report-template.html
```diff
@@ -15,11 +15,267 @@
...
+                  }
+                }
+                if (coverage_operator === '='){
+                  if(percentage === coverage){
+                    array.push('^'+d+'$');
+                  }
+                }
+              });
+              array = array.join('|');
+              column.search(array, true, false, true).draw();
```
mglb, line 137: > When a filter is active, if we change the value of any dropdown and click on apply, it currently filters from the previous output (i.e, from the dataTable that we get after the first filter is applied)
>  I've tried adding this:
> `report = $('#report_table').DataTable(); report.search('').draw();`
> at the beginning of applyFilters() function to clear any current filter but it didn't seem to make any change

This code applies filter to each column, is this correct? If so, I guess you have to clear search for each column too.

mglb (commented):
Avoid giving `id`s with numbered suffixes to everything. Use `id` only for specific parts of the page (e.g. sections) if needed. Organize elements into components, specify their type using tag and class names, and refer to them using relations, like `>` selector (for unique children, like "operator" drop down inside filter entry) or child index (for filter entries).

You can use tag's custom attributes (`data-...`) to attach custom data if needed. Optionally look into newer technologies like web components, templates, shadow DOM, etc.

Example element organization:

```html
<section id="filter">
    <ul class="filter-entries">
        <li data-uid="1">
            <select class="filter-entry-type" onchange="entry_changed(this)">...</select>
            <select class="filter-entry-operator" onchange="entry_changed(this)">...</select>
            <select class="filter-entry-value" onchange="entry_changed(this)">...</select>
            <button class="filter-entry-remove" onclick="remove_entry(this.parentElement)">...</button>
        </li>
        <li data-uid="2">...</li>
        <li data-uid="3">...</li>
        ...
    </ul>
    <div class="filter-actions">
        <button class="filter-apply">Apply</button>
        <button class="filter-clear">Clear</button>
    </div>
</section>
```

With such organization element removal can be shortened to something like:

```javascript
function remove_entry(element) {
    const uid = parseInt(element.dataset.uid);
    // use uid to cleanup internal structures if needed
    element.remove();
}
```
---

Instead of storing filter in separate array, you can build it from DOM (i.e. iterate `.filter-entries` and read type/operator/value fields for each element) when "Apply" is clicked. This will allow to avoid synchronization between user choices and structures in JavaScript (and will shorten `remove_entry` from snipped above to just one line).

---

Instead of handling every mode independently in multiple places, create an array describing every mode. Something like:

```javascript
const filter_types = {
    coverage: {
        name: "Coverage",
        operators: ["<", "<=", ">", ">="],
        value_field_factory: function() {
            // create and return HTML element that shoud appear as a value field for this filter type
            // call it when mode/filter type changes and value field must be replaced.
        },
        extract_value: function(value_element) {
            // extract entry value from entry_element and return it
            return parseInt(value_element.value);
        },
        apply: function(entry_element) {
            // apply filter entry of this type
        }
    }
};
```

---

Bugs/things that can be changed later:

* Use `<input type="number">` for percentages
* The first sticky column's width changes depending on content. This causes visual detachment of the second sticky column. You either need to make first column's width constant, or change th esecond column's position when the first one is resized. The latter solution is better.
Screenshot 2021-06-23 at 11-17-52 SystemVerilog Report

mglb:
Also - there are spaces in empty lines in conf/report/report-template.html

Prahitha:
> Also - there are spaces in empty lines in conf/report/report-template.html

Ah right, sorry about that. I'll fix it and work on the suggestions. Thank you for the feedback!

mglb (commented):
Much cleaner!

Hint: instead of `"something=" + some_var + " x=" + other_var` you can type `` `something=${some_var} x=${other_var}` `` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

Prahitha:
Still figuring out how to implement AND and OR dynamically. For consistency, everything is AND now. Any feedback is appreciated, thank you!

mglb (changes requested):
>     * Still figuring out how to implement AND and OR dynamically. For consistency, everything is AND now

Changes suggested in inline comments below make it a lot easier - you'll need to change filter callbacks:

* `value_field_factory` could use value-reading-callback to get current values (if they exist) and add more/remove existing fields when needed. `onchange` event of every field should call this callback again. Example:
  * Initially (no existing values), it would create one field more than it does now:
    ```
    [ coverage     ▾] [ <=  ▾] [ 30   ▾]  [ ... ▾]
    ```
    The new field would offer values ""/"AND"/"OR".
  * When 4th field has changed and it is not empty, it should create additional set of value fields:
    ```
    [ coverage     ▾] [ <=  ▾] [ 30   ▾]  [ OR ▾]  [ .. ▾] [ ...  ▾]  [ ... ▾]
    ```
  * When 4th field is changed back to empty, remove value fields it controls. To make things easier, just group AND/OR field + fields it controls using `<span>`:
    ```
                      ┌───── span ─────┐  ┌ span ┐
    [ coverage     ▾] [ <=  ▾] [ 30   ▾]  [ ... ▾]

                      ┌───── span ─────┐  ┌───────── span ─────────┐  ┌ span ┐
    [ coverage     ▾] [ <=  ▾] [ 30   ▾]  [ OR ▾]  [ .. ▾] [ ...  ▾]  [ ... ▾]
    ```
  * Make "4th field" check in a loop to handle more than two sets of AND/OR fields & values.

* value-reading-callback (proposed in comments), instead of `{operator: ..., value: ...}`, could read/return `[ {relation: ..., operator: ..., value: ...}, {relation: ..., operator: ..., value: ...}, ...]`.

Note that the example above shows coverage field, but I think it would be actually better to set the coverage value using simple range: `Coverage:  from [0  ▾]% to [100 ▾]%`.
Or it could be a range + AND/OR ;)
In any case, let's start with implementation of AND/OR on tool/test type fields.

conf/report/report-template.html
```diff
@@ -74,64 +86,75 @@
...
                   }
                 }
               });
             }
           });
           values = [...values];
           values = values.join('|');
-          table.search(values, true, false, true).draw();
+          if (tool_selected === true)
+            table.columns(tool_column).search(values, true, false, true).draw();
```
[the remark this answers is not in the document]
Prahitha: Ah right, that should be tool_array

conf/report/report-template.html
```diff
@@ -33,9 +38,250 @@
...
+          span_id = parseInt(span_id)
+          if ($(`li[data-uid=${id}] span[data-uid=${span_id+1}]`).length > 0)
+            $(`li[data-uid=${id}] span[data-uid=${span_id+1}]`).remove()
+          var iter = $(`li[data-uid=${id}] span`).length
+          if (iter === 0)
+            var filter_condition = $(`<span data-uid=${iter}></span>`).insertAfter(`li[data-uid=${id}] .filter-entry-type`)
+          else
+            var filter_condition = $(`<span data-uid=${iter}></span>`).insertAfter(`li[data-uid=${id}] span[data-uid=${iter-1}]`)
+          var filter_entry_operator =  $('<select class="filter-entry-operator"><option value=""></option></select>').appendTo(`li[data-uid=${id}] span[data-uid=${iter}]`)
+          var filter_entry_value =  $('<select class="filter-entry-value"><option value=""></option></select>').appendTo(`li[data-uid=${id}] span[data-uid=${iter}]`)
```
Prahitha, line 70: this seemed to repeat across the `tool` and `type` filters too in the `value_field_factory`. would it make sense to write a function to remove the repetition?

Prahitha:
I wasn't sure how AND works for the tool so I implemented only OR for now. what should `is slang AND is surelog` supposed to return?

conf/report/report-template.html
mglb: Sure

mglb (commented):
I see you did something like `git merge master` on your PR branch. This effectively put all new commits from the `master` branch between your commits. This can cause some problems:
* When your PR is merged, all new commits from your branch are appended on top of the `master` branch, so the commits order is changed. This can lead to conflicts.
* You've put your recent changes on the merge commit, mixing them with all new changes from the master branch.
* It makes it slightly harder to browse in some tools.

`git rebase master` would be better choice - it takes the master branch as-is and puts unique commits from your branch on top of it.

Also, it seems like you've fixed merge conflicts semi-automatically (which e.g. removed filtering UI by accident).

To fix the branch, read about git rebase and resolving merge/rebase conflicts, then do something like:

```
# make a backup
git checkout -B filter-ui-backup
# go back to filter-ui branch
git checkout -
git reset --hard 96aaca6 # go back to last commit before git merge ("fixed function name")
git fetch origin master

# Rebase the master branch
# Please read a bit about rebasing and merge conflicts, as you'll need to resolve a few conflicts
git rebase origin/master
# ... fix conflicts, finish rebasing ...

# Import your code from your recent commit ("or for tool").
git cherry-pick -m 1 4cde5ba
# ... fix conflicts, add fixed files ...
git cherry-pick --continue
```

You can always restore the backup with:
```
git reset --hard filter-ui-backup
```


> I wasn't sure how AND works for the tool so I implemented only OR for now. what should is slang AND is surelog return?

Good question. Maybe it would be better to always use OR (and replace drop down list with regular "or" text). It could be further simplified by removing all but first "is/is not" fields.
It would end up looking like this:

```
[ Tool   ▾][ is   ▾][            ▾]

# Picking "SomeTool":

[ Tool   ▾][ is   ▾][ SomeTool   ▾] or [           ▾]

# Picking "OtherTool":

[ Tool   ▾][ is   ▾][ SomeTool   ▾] or [ OtherTool  ▾] or [            ▾]

# Removing "OtherTool" by picking empty item:

[ Tool   ▾][ is   ▾][ SomeTool   ▾] or [           ▾]
```
There's also [multi-choice list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#attr-multiple), but it doesn't show as a drop-down list.

Another way is to use AND/OR to select how other filters work, e.g.:
* Coverage: > 50; Tool: X or Y → show test where either tool X or tool Y have coverage > 50
* Coverage: > 50; Tool: X and Y → show test where both tool X and tool Y have coverage > 50

Would it be hard to do?

Prahitha:
I tried implementing both `coverage: > 50; Tool: X or Y` and `coverage: > 50; Tool: X and Y` - and I think it's doable (though the logic/code seems kind of messy now). But if we go ahead with this, should we give an alert when someone tries giving only `tool: X and Y` without test type or coverage? Or is that case something we're not interested in?

(also, thank you for helping with the rebase/merge conflicts. I hope I did it right - is there a way I could check without pushing to my repo? what is the right way to pull changes from the upstream? just wanted to know because I did this while pulling the changes from my previously merged PR - footer detail - since it had some changes in the report-template.html)

mglb:
> I tried implementing both coverage: > 50; Tool: X or Y and coverage: > 50; Tool: X and Y - and I think it's doable (though the logic/code seems kind of messy now). But if we go ahead with this, should we give an alert when someone tries giving only tool: X and Y without test type or coverage? Or is that case something we're not interested in?

IMO alert is unnecessary - this would be equivalent of something like "tests executed by X (and|or) by Y", which currently gives the same results for both logic operators.

BTW. can you write a short summary (in a form of a list or something) how the filter is intended to work? It will be useful in spotting remaining issues. In the end it would be necessary anyway to be put in a documentation or a code comment.

> is there a way I could check without pushing to my repo?

Since `git clone` accepts paths to directories as an URI, you can clone your local repo in another directory (`git clone C:/develop/my_project -b filter-ui`) for a fully clean check.
Usually it is enough to skip cloning and just check that everything is commited (`git status`), then just do regular build/test.

> what is the right way to pull changes from the upstream?

`git pull --rebase origin master` (you can skip `origin master` when your branch has upstream reference set) which does `git fetch origin master; git rebase origin/master`.
As for resolving conflicts, there are multiple tools that help with it. I recommend your regular text editor, at least at the beginning - just to know how it really works, so you won't accidentally commit "conflict markers" like `=======` and will know what's going on when you'll see them.

Prahitha:
I'm not sure how to extend/make this more robust, the code could also be cleaner I guess?

What I tried doing is:
For tool: Whatever the condition is (AND/OR), the specified tool columns will be shown.
For coverage: If tool has 'AND', coverage is shown such that all the tools satisfy the condition. If tool has 'OR', coverage is shown such that at least one of the tools satisfies the condition. If there is no tool filter, I wasn't sure how only coverage 'OR' would be useful. If there is 'AND' for coverage, I passed both the operators and the values together and filtered simultaneously.
For type: As of now, it doesn't matter if tool has 'AND'/'OR' since the tool type is added as a classname to the row. Filtering for test type with 'AND'/'OR' is similar to how it's done with coverage.

(Also, I think I messed up the rebasing. I checked the log in the console and it seemed fine. I'll try it again)

conf/report/report-template.html
```diff
@@ -42,183 +43,363 @@
...
-        tool_index = applied_filters.indexOf("tool");
-        applied_filters = arrayMove(applied_filters, tool_index, 0);
-        applied_filters = arrayMove(applied_filters, tool_index+1, 1);
-        applied_filters = arrayMove(applied_filters, tool_index+2, 2);
+      applied_filters.sort((a, b) => (a.priority < b.priority) ? 1 : -1);
+      for(const key in applied_filters){
+        if (applied_filters[key]['filter'] === 'tool')
+          tool_relation = applied_filters[key][0]['relation'];
+        if (applied_filters[key]['filter'] !== 'tool' && (Object.keys(applied_filters[key]).length-2 === 2))
+          filter_types[applied_filters[key]['filter']].apply(applied_filters[key][0]['operator'], applied_filters[key][0]['value'], applied_filters[key][1]['operator'], applied_filters[key][1]['value'], applied_filters[key][0]['relation']);
```
Prahitha, line 394: I thought that there wouldn't be more than 2 operators/values in the case of coverage and type and implemented it this way. Please let me know if there is a better way to do this.

mglb: Didn't check how the whole code works, but generally:
* Limit visible/analyzed columns to all columns from the tool filter (relation doesn't matter here)
* For each row in every visible column apply a coverage/type filter. Store the results in a 2d array.
* Apply the tool filter (relations) to every row in the 2d array generated above. Show only rows that pass this filter.

As for variable length filters - create a callback function that can be passed to `Array.reduce()` function. This callback would take accumulated value and a next operator/value pair, and return `accumulator (operator) next_value`. So basically something you already have, but run it in a loop and assign result of each step to "first value" of the next step.

mglb (commented):
> What I tried doing is:
> For tool: Whatever the condition is (AND/OR), the specified tool columns will be shown.

Seems like when the filter is applied and then changed, the new filter is appended to previously applied one.

Steps I've performed:
* tool is icarus (apply) → OK
* tool is icarus AND is not moore (apply) → OK
* tool is icarus AND is moore (apply) → Wrong - all columns are shown
* tool is icarus AND is not moore (apply) → Still wrong - all columns
* tool is icarus (apply) → Still wrong

More visible with following steps:

* tool is icarus (apply) → OK
* tool is icarus AND is moore (apply) → OK
* tool is icarus AND is moore_parse (apply) → Wrong, table shows icarus, moore, moore_parse
* (Remove all filters) → OK
* tool is icarus (apply) → Wrong, table shows icarus, moore, moore_parse

> For coverage: If tool has 'AND', coverage is shown such that all the tools satisfy the condition. If tool has 'OR', coverage is shown such that at least one of the tools satisfies the condition. If there is no tool filter, I wasn't sure how only coverage 'OR' would be useful. If there is 'AND' for coverage, I passed both the operators and the values together and filtered simultaneously.

Seems to work.

> For type: As of now, it doesn't matter if tool has 'AND'/'OR' since the tool type is added as a classname to the row. Filtering for test type with 'AND'/'OR' is similar to how it's done with coverage.

"Type is preprocessing" gives "no matches found", but there are preprocessing tests (e.g. "Whitespace").


Minor problems:
* When AND is switched to OR (or the other way around), fields following it are cleared. They should be cleared only when AND/OR is changed to empty value. Or even better - they should be cleared/removed only when "Apply" is clicked. This way accidental selection of empty value (which can happen by pressing up/down key or scrolling in wrong moment) won't remove user choices.
* Applying a tool or type filter breaks layout a bit (horizontal scrollbar appears in firefox; second and first column become separated in chrome). "Remove all filters" doesn't fix it. Coverage filter alone seems to not trigger this.

Screenshot_20210730_130551
Screenshot_20210730_130517

tgorochowik (approved):
Let's merge this :tada:

PR #1640, opened 2021-07-15
Add daily builds

tgorochowik:


tgorochowik:
@wsnyder you might be interested in this :)

Now that we use github actions exclusively this should work fine

PR #1643, opened 2021-07-16
VeribleExtractor: Set correct file list root

mglb:
Sets correct file list root and makes sure all paths in the file list are absolute.


tgorochowik (approved):
LGTM, thanks!

PR #1655, opened 2021-07-17
footer detail

### LINKED ISSUE #1641: Improve report footer
Currently, we have the following footer:
```
Generated using sv-tests, revision: fba2e1b.
```

It used to be enough, but now with daily builds, the sha of the revision is not an an ideal identification. Now the report can (and will) be deployed multiple time from a single commit.

We need to add more info to the footer, I think adding github actions job id and the date (or date and time) of the build would be nice.

```
Generated using sv-tests, revision: fba2e1b, build_id: 123456789, date: 2021-07-15 14:15:16.
```

The build ID should be available in the environment (https://docs.github.com/en/actions/reference/environment-variables). I think `GITHUB_ACTION` is what we need.

It would be nice to make the build id clickable, the URL could point to: https://github.com/chipsalliance/sv-tests/actions/runs/1030721027 (with the correct id at the end).

we also need to take local builds into account. If `GITHUB_ACTION` is not available in the environment, we could put static text instead, e.g.:

```
Generated using sv-tests, revision: fba2e1b, build_id: local, date: 2021-07-15 14:15:16.
```

In this case the build_id would not be clickable.

Prahitha:
Should fix #1641 
From the feedback earlier:
- the datetime can be generated in Python
- the report-template should probably use the passed build_id from Python for checking the ID

### DIFF
--- a/.github/workflows/sv-tests-ci.yml
+++ b/.github/workflows/sv-tests-ci.yml
@@ -73,10 +73,19 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           path: ./out/
+      - name: Get current time
+        uses: 1466587594/get-current-time@v2
+        id: current-time
+        with:
+          format: YYYYMMDD-HH
       - name: Summary
         run: |
           ./.github/workflows/summary.sh
           ./.github/workflows/report.sh
+        env:
+          global:
+            - BUILD_ID = ${{GITHUB_ACTIONS}}
+            - DATETIME = "${{ steps.current-time.outputs.readableTime }}"
       - name: Update sv-tests-results repository
         if: github.ref == 'refs/heads/master'
         run: |
--- a/.github/workflows/sv-tests-ci.yml
+++ b/.github/workflows/sv-tests-ci.yml
@@ -73,19 +73,10 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           path: ./out/
-      - name: Get current time
-        uses: 1466587594/get-current-time@v2
-        id: current-time
-        with:
-          format: YYYYMMDD-HH
       - name: Summary
         run: |
           ./.github/workflows/summary.sh
           ./.github/workflows/report.sh
-        env:
-          global:
-            - BUILD_ID = ${{GITHUB_ACTIONS}}
-            - DATETIME = "${{ steps.current-time.outputs.readableTime }}"
       - name: Update sv-tests-results repository
         if: github.ref == 'refs/heads/master'
         run: |
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -20,6 +20,10 @@
   </head>
 
   <script>
+    if (process.env.BUILD_ID !== 'local') 
+      $('.addLink').attr('href', `https://github.com/chipsalliance/sv-tests/actions/runs/${process.env.BUILD_ID}`);
+    else
+      $('.addLink').removeAttr('href');
     $(document).ready( function () {
       $('#report_table').DataTable( {
         paging: false,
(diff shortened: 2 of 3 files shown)


.github/workflows/sv-tests-ci.yml
```diff
@@ -73,10 +73,19 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           path: ./out/
+      - name: Get current time
+        uses: 1466587594/get-current-time@v2
+        id: current-time
+        with:
+          format: YYYYMMDD-HH
```
tgorochowik, line 80: It should be okay to always read it in the python script

.github/workflows/sv-tests-ci.yml
```diff
@@ -73,10 +73,19 @@ jobs:
...
       - name: Summary
         run: |
           ./.github/workflows/summary.sh
           ./.github/workflows/report.sh
+        env:
+          global:
+            - BUILD_ID = ${{GITHUB_ACTIONS}}
+            - DATETIME = "${{ steps.current-time.outputs.readableTime }}"
```
tgorochowik, line 88: These changes should not be needed, GITHUB ACTIONS is already an environment variable, no need to add it here

conf/report/report-template.html
```diff
@@ -20,6 +20,10 @@
   </head>
 
   <script>
+    if (process.env.BUILD_ID !== 'local')
```
tgorochowik, line 23: This should operate on jinja2 template arguments, this will attempt to get the env of the user browser, we need to read those values from the environment of the build machine

tools/sv-report
```diff
@@ -557,13 +557,24 @@ try:
...
     with open(args.out, 'w') as f:
         f.write(
             report.render(
                 report=data,
                 database=database,
                 database_urls=urls,
-                revision=args.revision))
+                revision=args.revision,
+                build_id=os.environ['BUILD_ID'],
+                datetime=os.environ['DATETIME']))
```
tgorochowik, line 577: So with those added, you should be able to use `{{build_id}}`  and `{{datetime}}` in the js file, no need to try to access env there

tgorochowik (changes requested):
Thanks for working on this PR! This is going in the right direction but needs some more work for it to work

tools/sv-report
```diff
@@ -557,13 +557,21 @@ try:
         report = jinja2.Template(
             f.read(), trim_blocks=True, lstrip_blocks=True)
 
+    try:
+        build_id = os.environ['GITHUB_ACTION']
+    except:
+        build_id = 'local'
```
tgorochowik, line 563: We could covere it to

`os.environ.get('GITHUB_ACTION', 'local')`

The second parameter is the default value if the env variable is not found, this will make it a little bit simpler

tools/sv-report
```diff
@@ -557,13 +557,21 @@ try:
...
+    datetime = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
tgorochowik, line 564: I think it would be better to come up with a different variable name - right now this overwrites the module (the name of the veritable is the same as the module).

Maybe let's change it to `build_time` / `build_datetime`?

tgorochowik:
The CI failures also need to be checked, it seems they all fall with:
```
CRITICAL| Unable to generate report: expected token 'end of statement block', got ':'
```

conf/report/report-template.html
```diff
@@ -143,7 +143,14 @@ <h1 class="headline"><a href="https://github.com/chipsalliance/sv-tests">SV-Test
   <br> <br>
   <footer id="footer">
     Generated using <a href="https://github.com/chipsalliance/sv-tests">sv-tests</a>,
-    revision: <a href="https://github.com/chipsalliance/sv-tests/commit/{{revision}}">{{revision}}</a>.
+    revision: <a href="https://github.com/chipsalliance/sv-tests/commit/{{revision}}">{{revision}}</a>,
+    {% if build_id != 'local' %}
+      {% set href = https://github.com/chipsalliance/sv-tests/actions/runs/ + build_id %}
```
tgorochowik, line 148: This is the candidate for the cause of the CI failures, I don't think we can add strings and variables like this in jinja2 templates.

Maybe the extra href variable is not needed and everything can be put together in the line below?

tgorochowik (approved):
Great, thanks for working in this!

PR #1658, opened 2021-07-19
Fix event_sequence test

MikePopoloski:
Assigning to nets from a procedural context is not allowed, so make them variables instead.

Example error:
```
../..tests/chapter-9/9.4.2.4--event_sequence.sv:30:10: error: cannot assign to a net within a procedural context
				@seq y = 1;
				     ^
```


tgorochowik (approved):

tgorochowik:
Thanks!

PR #1677, opened 2021-08-04
Add history graph util

### LINKED ISSUE #1636: History graph
Add a small graph with history to the sv tests report.

Some ideas:
* only store the total number of test passing for each tool (or maybe percentage to avoid thinking something is better when there are more tests?)
* append a line with all the totals to some .csv file stored in the results repository
* always truncate that to like 100 entries
* draw a graph on the bottom of sv-tests

CC @hzeller

tgorochowik:
Closes #1636

This adds a simple history graph (generated using commits from the sv-tests-results).

Currently it looks like this:

It uses a MIT licensed https://www.chartjs.org/ library.

TODO:

- [x] Add a CI job that generates it
- [x] Publish it next to the original report
- [ ] Add links in the report to the history graph

### DIFF
--- a/tools/history-graph
+++ b/tools/history-graph
@@ -15,6 +15,7 @@ import sys
 import jinja2
 import datetime
 import argparse
+import binascii
 
 from git import Repo
 from random import randint
@@ -91,8 +92,8 @@ for tool in datasets:
 # randomize colors
 colors = dict()
 for tool in datasets:
-    r, g, b = randint(0, 255), randint(0, 255), randint(0, 255)
-    colors[tool] = '#%02x%02x%02x' % (r, g, b)
+    color = hex(binascii.crc32(tool.encode('utf-8')))[-6:]
+    colors[tool] = f'#{color:0<6.6}'
 
 try:
     with open(args.template) as f:
--- a/tools/history-graph
+++ b/tools/history-graph
@@ -54,15 +54,10 @@ args = parser.parse_args()
 
 def single_report_data(index, csv_file, data):
     with io.BytesIO(csv_file.data_stream.read()) as f:
-        reader = csv.reader(io.StringIO(f.read().decode('utf-8')))
+        reader = csv.DictReader(io.StringIO(f.read().decode('utf-8')))
         for r in reader:
-            # skip header
-            if r[1] == 'Tool':
-                continue
-            if 'False' in r[2]:
-                data[r[1]][index]['fail'] += 1
-            else:
-                data[r[1]][index]['pass'] += 1
+            key = 'pass' if r['Pass'] == 'True' else 'fail'
+            data[r['Tool']][index][key] += 1
 
 
 repo = Repo(args.results_dir)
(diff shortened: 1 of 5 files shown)


tgorochowik:
CC @hzeller

mithro:
Pretty!

tools/history-graph
```diff
@@ -0,0 +1,114 @@
...
+
+# convert lists to strings
+labels = ', '.join(f'"{x}"' for x in data[next(iter(data))].keys())
+
+for tool in datasets:
+    datasets[tool] = ', '.join(str(x) for x in datasets[tool])
+
+# randomize colors
+colors = dict()
+for tool in datasets:
```
hzeller, line 98: mmh, that means we generate a different color for a particular tool each day ? That will probably be confusing for someone checking the same tool and just looking for the same color.

To make the colors repeatable, I'd base the 'random value' on the hash() name of the tool. Thus, the same tool always has the same repeatable random value and does not need any special configuration for that.

hzeller (commented):
Very nice looking!

tools/history-graph
hzeller: I think `hash()` is not stable between invocations and thus not useful here, so choosing some stable function from hashlib is probably needed.

tools/history-graph
```diff
@@ -0,0 +1,114 @@
...
+
+
+def single_report_data(index, csv_file, data):
+    with io.BytesIO(csv_file.data_stream.read()) as f:
+        reader = csv.reader(io.StringIO(f.read().decode('utf-8')))
+        for r in reader:
+            # skip header
+            if r[1] == 'Tool':
+                continue
+            if 'False' in r[2]:
```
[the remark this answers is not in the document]
tgorochowik: I changed the `csv` reader to one based on dicts, therefore we no longer have to use numeric indexing - this improves the readability a little bit.

In addition I put key extraction to a line above (similar to what you suggested) and then do the incrementation in a separate line. This seems to be the most readable for me.

tools/history-graph
hzeller: :+1:

tools/history-graph
```diff
@@ -0,0 +1,110 @@
...
+# parse args
+args = parser.parse_args()
+
+
+def single_report_data(index, csv_file, data):
+    with io.BytesIO(csv_file.data_stream.read()) as f:
+        reader = csv.DictReader(io.StringIO(f.read().decode('utf-8')))
+        for r in reader:
+            key = 'pass' if r['Pass'] == 'True' else 'fail'
+            data[r['Tool']][index][key] += 1
```
hzeller, line 61: Nice, much easier to follow!

hzeller (approved):
Very nice.

This will show up as a separate output, so we should have a link in the README where this will be generated. Later we then can also refine that with making a 'tabbed' view with a small (possibly just simply static snapshot) image or something so that we make it easy to discover.

PR #1687, opened 2021-08-11
Don't upload all the output for the PR commenter

tgorochowik:
It does not need the whole report, it is faster and saves us some quota
for the remaining artifacts where they are actually used.


tgorochowik:
The PR comment job is expected to fail here - we will upload a different set of artifacts than the commenter job (which runs in the context of the main branch to be able to access the token).

PR #1690, opened 2021-08-13
History graph improvements

tgorochowik:
Changes:
* Added proper labels to x and y axis
* Changed the graph type to stepped (no diagonal lines, which suggest "guessing" what happened in between builds)
* When hovering the cursor over a tool name in the legend, all other tools are faded out - this makes it easier to identify the graph for a given tool, example below

### DIFF
--- a/conf/report/history-graph-template.html
+++ b/conf/report/history-graph-template.html
@@ -53,4 +53,3 @@
     </script>
 </body>
 </html>
-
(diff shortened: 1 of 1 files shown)


hzeller (approved):
Very nice, this is a good usability improvement!

(Speaking of labels: what is their current sorting ? It looks like they come straight out of some hash-table iterator; displaying them case-ignore alphabetically would make it easier to find)

tgorochowik:
Yes, they are currently pretty much random, that's a good point! Let's change that in a separate PR though

PR #1703, opened 2021-08-17
sv2v runner: support defines

zachjs:
Similar logic already exists in other runners. Relevant tests which previously failed now appear to pass.


tgorochowik (approved):
Thanks!

PR #1705, opened 2021-08-18
Make Yosys, Yosys-SV and UHDM-Yosys runners equivalent

kbieganski:
Currently the runner for UHDM-Yosys runs more commands than the ones for vanilla Yosys and Yosys-SV, which results in an unfair comparison. This PR adds the missing commands to the Yosys and Yosys-SV runners. It also removes the `-check` flag from the `hierarchy` command, as it would cause some tests to fail (like this one: https://chipsalliance.github.io/sv-tests-results/#UhdmYosys|utd-sv|utd-sv_bw_dtl_impctl_pulldown).

It also cleans up the UHDM-Yosys runner a bit (imports, unused vars).

### DIFF
--- a/tools/runners/UhdmYosys.py
+++ b/tools/runners/UhdmYosys.py
@@ -10,7 +10,7 @@
 # SPDX-License-Identifier: ISC
 
 import os
-import shutil
+import sys
 
 from BaseRunner import BaseRunner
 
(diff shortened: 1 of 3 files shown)


tgorochowik (approved):
LGTM!

PR #1708, opened 2021-08-19
history: take not yet committed changes into account

### LINKED ISSUE #1707: history graph: take current build into account
The history util takes the csv files from git history and uses them to generate a graph.

It does not take the _current_ (pre commit) csv into account. We have to add special handling for that otherwise we will always miss the current state in the graphs.

tgorochowik:
When generating the report and the history graph, current results are
not yet committed. We have to handle them separately.

fixes #1707

### DIFF
--- a/tools/history-graph
+++ b/tools/history-graph
@@ -10,6 +10,7 @@
 # SPDX-License-Identifier: ISC
 
 import io
+import os
 import csv
 import sys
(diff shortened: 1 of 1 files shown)


tgorochowik:
This is not a controversial change and fixes an annoying bug, let's merge this.

PR #1709, opened 2021-08-19
Make PR artifacts more useful

tgorochowik:
Keep data about new failures, new passes and full csv files that were
used for comparison.

### DIFF
--- a/.github/workflows/comment-pr.yml
+++ b/.github/workflows/comment-pr.yml
@@ -13,6 +13,7 @@ jobs:
     if: ${{ github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' }}
     steps:
       - name: 'Download artifact'
+        id: get-artifacts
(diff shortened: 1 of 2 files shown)


hzeller:
Are these the ones the "compared test results" table is build from ? If so, could we embed links in the table to the artifacts that can further explain what was happening ?

tgorochowik:
Yes, the files are the original csv files (before pr and after pr) and in addition to that, pre filtered csv files that only show new failures and new passes

hzeller:
Nice, so could we craft URLs to this information in the generated pull request comment that contains the results table ?

tgorochowik:
Converting to draft as adding the url requires some experimentation.

tgorochowik:
Oh, right. We won't be able to test it here as the workflow that posts the comment has to be already on the main branch. Let me add some logs to CI that generate the link and test it there. If the URL is generated correctly, then I will add the code that appends it to the comment separately.

PR #1710, opened 2021-08-20
artifacts URL tests

tgorochowik:


tgorochowik:
Should be good to go, let's merge it.

PR #1718, opened 2021-08-21
Fix mailbox test variable declarations

jrudess:
The mailbox tests have a syntax problem where variables are defined after functional logic in the initial block.   Slang detects this and gives a valid error. The fix is to move the mailbox creation to be after the variable declarations.

```
tests/chapter-15/15.4--mailbox-blocking.sv:21:2: error: declaration must come before all statements in the block
        string msg = "abc";
        ^
```


kbieganski (approved):
LGTM

PR #1719, opened 2021-08-21
Add tests/chapter-20/20.14--coverage.sv

jrudess:
Adds a test to cover parsing of the system tasks and functions defined in chapter 20.14.  This test-case also includes the pre-defined coverage macros defined in 40.3.1.


tgorochowik (approved):
Sorry for a delayed reply, we somehow missed this PR.

Thanks for the contribution!

PR #1721, opened 2021-08-23
fixed bug in type filter

Prahitha:

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -399,7 +399,7 @@
       </thead>
         {% for tag, info in database.items() %}
         {% for tool, tooldata in report|dictsort %}
-         {% if loop.first %}
+         {% if loop.last %}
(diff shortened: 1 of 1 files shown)


mglb:
Problem: type filter set to "preprocessing" does not find anything

Problem analysis:
Some tools don't support preprocessing tests. This is indicated by empty (gray) cells in the table (see tests 22.*).
The data from a specific tool (first one) in each row is used to assign the test types to the whole row. This happens to be Icarus, which does not support preprocessing tests and its test data is empty in such tests.

The fix:
The patch uses the last tool's data for marking a row. This depends on the fact that the last tool supports all test types.

Ideally it should collect the types from all tools. I guess jinja2's `set` command would be useful.

Prahitha:
Thank you for adding the description! I tried using `set` to concat the list from each tool in a row into a single list/set. However, this prints `None` after each iteration - even when the append works as expected. I'm now trying to send the data in the required format from the `sv-report` file to the jinja template.

PR #1733, opened 2021-09-01
and/or between filters

Prahitha:
The functionality is kind of limited and still needs some work. For example, it assumes that the user will not enter: coverage > 30 AND/OR coverage > 40 AND/OR coverage > 50

I would also like to know if the data structure used here is okay or if I should implement an array with priority like earlier. Any feedback is appreciated. Thank you!

### DIFF
--- a/conf/report/report-template.html
+++ b/conf/report/report-template.html
@@ -49,9 +49,14 @@
       tool_array = table.columns()[0];
       tool_array = tool_array.splice(2);
       col_array = table.columns()[0];
+      var filters = $('<details><summary>Show advanced filters (experimental)</summary></details>').insertAfter(".dataTables_filter");
+      var section = $('<section id="filter"><ul class="filter-entries"></ul></section>').insertAfter("summary");
+      var global_relation = $('<select class="global-relation"><option value=""></option><option value="and">AND</option><option value="or">OR</option></select>').insertBefore('.filter-entries');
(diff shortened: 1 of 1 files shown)


mglb:
This doesn't seem to work in basic cases:
* (empty/default value) tool is icarus -> full table is visible;
   removing the empty value and leaving either OR or AND as the default should remove this problem.
* (AND) tool is icarus -> OK
* (OR) tool is icarus -> all tools visible with "no matching records"

"Remove all filters" doesn't remove filter from the table. Only filter fields are removed.

Why did you remove support for relation operators in entries (rows)? It has been useful to e.g. search "coverage <= 0 OR >= 100" AND "tool is X".

Code structure seems to be OK at first glance - lets make it work well first and optionally do some small optimization/refactoring later.

Prahitha:
Thanks for the feedback, I'll work on those issues. Also, remove all filters works fine for me. It takes a awhile to redraw the table though.

> Why did you remove support for relation operators in entries (rows)? It has been useful to e.g. search "coverage <= 0 OR >= 100" AND "tool is X".
@tgorochowik suggested that we have only one filter per row. So we agreed to first implement AND/OR between rows as a global relation and then extend it to AND/OR between every row.

mglb:
> Also, remove all filters works fine for me. It takes a awhile to redraw the table though.

Case that don't work in Chrome and Firefox:
* Select OR
* Add
* Select: tool is icarus
* Apply → "No matching records"
* Remove all filters → still "No matching records"

I did wait a few minutes for results.

Prahitha:
Ah yes, it doesn't seem to work when there are no matching records.

PR #1743, opened 2021-09-07
script to test filters

Prahitha:
The filters are being applied and the simulation, I think, works as expected but the results don't show up for some reason. It always shows "no matching records"

### DIFF
--- a/conf/report/testing.py
+++ b/conf/report/testing.py
@@ -1,0 +1,56 @@
+from selenium import webdriver
+from selenium.webdriver.support.ui import Select
+import sys
+
+def openPage(URL):
+    driver.get(URL)
+
+def addEntries(iter):
+    for _ in range(iter):
+        driver.find_element_by_class_name('filter-add').click()
+
+def fillEntry(num, type, operator, value):
+    entry_type = "//*[@class='filter-entry-type']"
+    select = Select(driver.find_elements_by_xpath(entry_type)[num])
+    select.select_by_value(type)
+    entry_operator = "//*[@class='filter-entry-operator']"
+    select = Select(driver.find_elements_by_xpath(entry_operator)[num])
+    select.select_by_value(operator)
+    entry_value = "//*[@class='filter-entry-value']"
+    select = Select(driver.find_elements_by_xpath(entry_value)[num])
+    select.select_by_value(value)
+
+def addGlobalOp(operator):
+    operator_type = "//*[@class='global-relation']"
+    select = Select(driver.find_element_by_xpath(operator_type))
+    select.select_by_value(operator)
+
+def applyFilters():
+    driver.find_element_by_class_name('filter-apply').click()
+
+def removeIcon(num):
+    driver.find_elements_by_class_name('filter-clear')[num].click()
+
+def removeAll():
+    driver.find_element_by_class_name('filter-remove').click()
+
+chromedriver_location = "D:\Downloads\chromedriver_latest\chromedriver.exe"
+driver = webdriver.Chrome(chromedriver_location)
+
+try:
+    URL = sys.argv[1]
+except:
+    URL = "D:/Outreachy/local-cp/sv-tests/out/report/index.html"
+
+openPage(URL)
+
+driver.implicitly_wait(10)
+driver.find_element_by_tag_name('details').click()
+addEntries(3)
+fillEntry(0, "coverage", "=", "50")
+fillEntry(1, "type", "is", "preprocessing")
+fillEntry(2, "tool", "is not", "icarus")
+addGlobalOp("and")
+applyFilters()
+removeIcon(2)
+removeAll()


mglb:
> works as expected but the results don't show up for some reason

The filter configuration set up by hand gives no results too, so that's not the test script fault.

Prahitha:
The ones set by hand in the Chrome started by Selenium/WebDriver don't work. However, if it's a normal Chrome tab that we open, it seems to work. For eg: coverage = 50 and type is preprocessing gives results in the normal Chrome tab but not by Selenium.

mglb:
This works for me (results in one row):
```
driver.implicitly_wait(10)
driver.find_element_by_tag_name('details').click()

addEntries(2)
fillEntry(0, "coverage", "=", "50")
fillEntry(1, "type", "is", "preprocessing")

addGlobalOp("and")
applyFilters()
```

Prahitha:
Ahh I'm not sure why it wasn't working earlier - I see 2 rows now. Also, it could be the "is not" that isn't working because it worked fine when I changed it to: 
`fillEntry(0, "coverage", ">", "50")
fillEntry(1, "type", "is", "simulation")
fillEntry(2, "tool", "is", "icarus")`

Prahitha:
Please let me know if I have to change something, thanks!

conf/report/testing.py
```diff
@@ -0,0 +1,74 @@
+from selenium import webdriver
+from selenium.webdriver.support.ui import Select
+import sys
+
+def setDriver(driver_location, driver=None):
```
mglb, line 5: `driver_location` has to be optional - in most cases it doesn't need to be provided to webdriver's constructor.
Alternatively you can leave only `driver=None` argument and use it for providing the driver object, so that `setDriver()` will set default one (Chromium), `setDriver(webdriver.Firefox())` will use Firefox, etc.

conf/report/testing.py
```diff
@@ -0,0 +1,74 @@
...
+    try:
+        URL = sys.argv[1]
+    except:
+        URL = "D:/Outreachy/local-cp/sv-tests/out/report/index.html"
+
+    driver = setDriver("D:/Downloads/chromedriver_latest/chromedriver.exe")
+    openPage(URL)
+
+    driver.implicitly_wait(10)
+    driver.find_element_by_tag_name('details').click()
```
mglb, line 66: This should be called automatically in one of the functions (`openPage` seems to be best choice).

conf/report/testing.py
```diff
@@ -0,0 +1,74 @@
...
+
+def removeAll():
+    driver.find_element_by_class_name('filter-remove').click()
+
+
+if __name__ == "__main__":
+    try:
+        URL = sys.argv[1]
+    except:
+        URL = "D:/Outreachy/local-cp/sv-tests/out/report/index.html"
```
mglb, line 60: This shouldn't be there. Requiring URL in cmd line without any fallback would be OK.

conf/report/testing.py
```diff
@@ -0,0 +1,74 @@
...
+
+    driver = setDriver("D:/Downloads/chromedriver_latest/chromedriver.exe")
```
mglb, line 62: This path shouldn't be there. It would be best to not set the driver's path at all. You can always do something like:
```
set PATH=%PATH%;C:\Downloads\chromedriver_latest\
python3 ./testing.py D:/Outreachy/local-cp/sv-tests/out/report/index.html
```
(this adds additional path just for the current terminal session)

conf/report/testing.py
```diff
@@ -0,0 +1,74 @@
...
+    driver.implicitly_wait(10)
+    driver.find_element_by_tag_name('details').click()
+    addEntries(3)
+    fillEntry(0, "coverage", "=", "50")
+    fillEntry(1, "type", "is", "preprocessing")
+    fillEntry(2, "tool", "is not", "icarus")
+    addGlobalOp("and")
+    applyFilters()
+    removeIcon(2)
+    removeAll()
```
mglb, line 74: Missing newline

mglb (changes requested):
Please rename the file to something less generic (filter_ui_test.py or something).

Please also fix Code Quality Check complaints. To fix them you can just copy shebang/license comment from any other python script in this repo. Just update copyright's year if it's not 2021.

mglb:
Oh, and one more thing: we should target already merged filter version (i.e. without global operator). Can you adapt it? Looks like the only required changes are `fillEntry` adjustment and removal of `addGlobalOp`.

conf/report/filter_ui_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+from selenium import webdriver
+from selenium.webdriver.support.ui import Select
+import sys
+
+def setDriver(driver=None):
+    if driver == None:
+        return webdriver.Chrome()
+    else:
+        return driver
```
mglb, line 20: This should set global variable `driver`. Otherwise there's no way to actually set the driver from another script.

conf/report/filter_ui_test.py
```diff
@@ -0,0 +1,74 @@
...
+def addEntries(iter):
+    for _ in range(iter):
+        driver.find_element_by_class_name('filter-add').click()
+
+def fillEntryType(num, type):
+    entry_type = "//*[@class='filter-entry-type']"
+    select = Select(driver.find_elements_by_xpath(entry_type)[num])
+    select.select_by_value(type)
+
+def fillSpan(num, operator, value, relation=None):
```
mglb, line 36: If you change it to `fillSpan(entry_id, *args)`, loop over `args` and assign each arg's value to a Select queried with an XPath like `f"(//*[@id='filter']/ul/li[{ entry_id + 1 }]/span/select)[{ loop_iteration + 1 }]"`, it will be possible to use it this way: `fillSpan(0, ">", "50", "or", "<", "80")`.

mglb (commented):
Just one small important change and one suggestion.

And please do `git rm conf/report/testing.py`

Prahitha:
Sorry it took me so long and thanks for being patient! (I was finishing up my mid-terms and some other stuff)

mglb:
LGTM, please make it "ready for a review" to allow merging.

tgorochowik:
How will that be used? Should we add a CI job for this?

mglb:
> How will that be used? Should we add a CI job for this?

This is a simple script to help developers/reviewers test the changes in filter UI. Instead of manually entering the filter settings after each small code change/full refresh, or describing what exactly did you click in what order to trigger some error, you can run/share a code like:

```python3
#!/usr/bin/env python3
import filter_ui_test as ft

ft.setDriver()
ft.openPage('https://chipsalliance.github.io/sv-tests-results/') # or URL to your local copy
ft.addEntries(1); ft.fillEntryType(0, 'coverage'); ft.fillSpan(0, '>', '50', 'and', '<=', '60')
ft.addEntries(1); ft.fillEntryType(1, "type");     ft.fillSpan(1, "is", "preprocessing")
ft.addEntries(1); ft.fillEntryType(2, "tool");     ft.fillSpan(2, "is", "icarus")
# Test: apply - remove one - apply
ft.applyFilters()
ft.removeIcon(2)
ft.applyFilters()
```

The result has to be manually verified, so it isn't suitable for CI in its current form.

conf/report/filter_ui_test.py
```diff
@@ -0,0 +1,67 @@
...
+    driver = setDriver(webdriver.Chrome())
+    openPage(URL)
+
+    addEntries(3)
+    fillEntryType(0, "coverage")
+    fillEntryType(1, "type")
+    fillEntryType(2, "tool")
+    fillSpan(0, ">", "50", "or", "<", "80")
+    fillSpan(1, "is", "preprocessing")
+    fillSpan(2, "is", "yosyssv", "and", "is", "surelog")
```
mglb, line 64: `yosyssv` is not currently present on https://chipsalliance.github.io/sv-tests-results/, so this example won't work with it. Please use e.g. "verible" and "verilator"

mglb (approved):

tgorochowik (commented):
Is `conf` the right directory for this script?

If so, can we make sure it is also taken into account when checking the formatting? see: https://github.com/chipsalliance/sv-tests/blob/master/Makefile#L145-L147

mglb:
> Is `conf` the right directory for this script?

It's quite good - it is near the stuff it is inteded to be used with. We can think about another directory (e.g. `development`) when we get more scripts like this one.

> If so, can we make sure it is also taken into account when checking the formatting? see: https://github.com/chipsalliance/sv-tests/blob/master/Makefile#L145-L147

@Prahitha please add `wildcard conf/report/*.py` in the Makefile linked above

mglb:
@Prahitha 
Just one more thing: please fix formatting errors (extra newlines: https://github.com/chipsalliance/sv-tests/pull/1743/checks?check_run_id=3805424990).

PR #1757, opened 2021-09-16
Fix randsequence tests

MikePopoloski:
Some of the randsequence tests referenced undeclared variables or had typos in their production names. This PR fixes them, and adds new tests that exercise the failing paths with a should_fail_because to make sure tools actually catch these kinds of errors since they are easy to make.

### DIFF
--- a/tests/chapter-18/18.17.2--if-else-production-statements_0_fail.sv
+++ b/tests/chapter-18/18.17.2--if-else-production-statements_0_fail.sv
@@ -8,18 +8,17 @@
 
 
 /*
-:name: if_else_production_statements_2
+:name: if_else_production_statements_0
 :description: randcase if-else test
+:should_fail_because: switch variable not declared
 :tags: 18.17.2
 */
 
 function int F();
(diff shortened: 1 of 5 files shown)


tests/chapter-18/18.17.2--if-else-production-statements_0_fail.sv
```diff
@@ -0,0 +1,24 @@
...
+// license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/ISC
+//
+// SPDX-License-Identifier: ISC
+
+
+/*
+:name: if_else_production_statements_0
+:description: randcase if-else test
+:should_fail_because: switch variable not declared
```
tgorochowik, line 13: These tests will not be green for tools that just parse and do no checking. I think it would be nice to have a separate test type (e.g. `elaboration`). This can replace the current `can_elaborate` flag.

Let's do this separately though (see: #1770)

tgorochowik (approved):
Thanks for the fixes!

PR #1773, opened 2021-09-22
Make elaboration a separate test type

tgorochowik:
WIP


tools/BaseRunner.py
```diff
@@ -48,7 +48,7 @@ def __init__(
             self,
             name,
             executable=None,
-            supported_features={'preprocessing', 'parsing', 'simulation'}):
+            supported_features={'preprocessing', 'parsing', 'elaboration'}):
         """Base runner class constructor
         Arguments:
         name -- runner name.
```
MikePopoloski, line 54: Probably the logic in get_mode needs to be updated too.

tools/runners/Slang.py
```diff
@@ -39,10 +41,9 @@ def prepare_run_cb(self, tmp_dir, params):
         for define in params['defines']:
             self.cmd.extend(['-D', define])
 
-        # hdlconv and utd tests are not semantically valid SystemVerilog, so we
-        # can only expect to run parsing successfully.
         tags = params["tags"]
-        if "hdlconv" in tags or "hdlconv_std2012" in tags or "hdlconv_std2017" in tags or "utd-sv" in tags:
+
+        if params['mode'] in ["parsing", "preprocessing"]:
```
MikePopoloski, line 46: --parse-only is mutually exclusive with -E (which is just a shortcut for --preprocess-only) which gets set above. I think some of the preprocessor tests aren't even valid syntactically and so can't be parsed.

MikePopoloski (commented):
I think the strategy should probably be to automatically find all tests that are currently marked with parsing or simulation and add 'elaboration' to them as well. That should preserve the current status-quo. Then once that's landed you can go back and remove the few tests that don't fit into the elaboration or simulation categories based on failures in the runners.

tools/BaseRunner.py
```diff
@@ -69,19 +69,16 @@ def get_mode(self, test_features, compatible_runners):
...
-        basic_features = ['parsing', 'preprocessing']
-        previous_required = False
-        for feature in basic_features:
-            if feature in test_features and feature not in self.supported_features and previous_required:
-                return None
-            if feature in test_features and feature in self.supported_features:
-                previous_required = True
 
-        features = ['simulation', *basic_features]
+        # select the first mode from the list that matches both the runner and
```
MikePopoloski, line 73: Nice, much simpler to reason about.

tools/runners/Slang.py
tgorochowik: Right, thanks!

tools/BaseRunner.py
```diff
@@ -69,19 +69,16 @@ def get_mode(self, test_features, compatible_runners):
...
+        # the test
+        modes_sorted = ['simulation', 'elboration', 'parsing', 'preprocessing']
```
MikePopoloski, line 75: Typo in "elboration"

tools/BaseRunner.py
tgorochowik: Right, I simplified it a little bit there

conf/fusesoc-configs/earlgrey.yml
```diff
@@ -17,4 +17,4 @@ conf_file: build/lowrisc_systems_top_earlgrey_0.1/sim-icarus/lowrisc_systems_top
 test_file: earlgrey.sv
 timeout: 360
 compatible-runners: all
-allow_elaboration: True
+type: parsing elaboration
```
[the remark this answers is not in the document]
tgorochowik: Hmm, but we should be able to elaborate it, the test case needs to be fixed then (however leaving it off here temporarily makes sense until we do fix it)

tgorochowik:
The changes in results are acceptable. Let's catch the remaining things that need to be addressed separately

MikePopoloski:
One thing I would be worried about is cases where slang (or similar tools) was previously doing elaboration successfully on a test and is now quietly just doing parsing. I think we should be sure we're doing the most involved form of testing that each particular test will support.

Also I would not have expected 180 test removals for slang -- I assume these are simulation tests but the vast majority of the existing tests marked as sim were mismarked and can be passed successfully with elaboration.

Of course both of these things are fixable in followup diffs, just wanted to point them out.

tgorochowik:
> One thing I would be worried about is cases where slang (or similar tools) was previously doing elaboration successfully on a test and is now quietly just doing parsing.

The majority of tests were converted to elaboration - elaboration is now the default mode if nothing else is specified. If somthing has been converted from elaboration to just parsing it was explicit (and was only done where you suggested it should be or when the runners explicitly marked them as parse only themselves - one of the things I wanted to do here is get rid of all of these hacks from runners). So hopefully we should be fine in this aspect.

> Also I would not have expected 180 test removals for slang -- I assume these are simulation tests but the vast majority of the existing tests marked as sim were mismarked and can be passed successfully with elaboration.

Yes, slang is currently marked to run preprocessing, parsing and elaboration, so unless we made a mistake somewhere, the removed tests should be simulation.

The simulation tests often check things that can only be verified at runtime (for example if 2+2 equals 4), that being said, even if it seems there is little benefit in checking those in elaboration tools, it might be that there is some weird corner case in one of the elaboration tools that one of this tests finds, so I will not oppose to individually marking those tests as elaboration too.
We have to doing manually though, as it is allowed to have a `should_fail` test that checks if 2+2 == 5 (that's also why we should add elaboration to the test type instead of allowing non-simulator tools to run the simulation tests)

MikePopoloski:
> The majority of tests were converted to elaboration - elaboration is now the default mode if nothing else is specified. If somthing has been converted from elaboration to just parsing it was explicit (and was only done where you suggested it should be or when the runners explicitly marked them as parse only themselves - one of the things I wanted to do here is get rid of all of these hacks from runners). So hopefully we should be fine in this aspect.

For example, it looks like a bunch of things under generators/templates/ are marked "simulation parsing" which means elaboration tools are now doing only the weaker parsing checking right? Maybe that folder was just missed in the automated update.

> The simulation tests often check things that can only be verified at runtime (for example if 2+2 equals 4), that being said, even if it seems there is little benefit in checking those in elaboration tools, it might be that there is some weird corner case in one of the elaboration tools that one of this tests finds, so I will not oppose to individually marking those tests as elaboration too.
We have to doing manually though, as it is allowed to have a should_fail test that checks if 2+2 == 5 (that's also why we should add elaboration to the test type instead of allowing non-simulator tools to run the simulation tests)

Agreed, if the test actually looks for runtime output then it should remain simulation-only. In the past when I've looked at some of the sim tests they didn't actually have any runtime output, but if all of the removed tests do then I'm fine with it.

tgorochowik:
> For example, it looks like a bunch of things under generators/templates/ are marked "simulation parsing" which means elaboration tools are now doing only the weaker parsing checking right? Maybe that folder was just missed in the automated update.

You are right, I missed those - thanks! As far as can tell those should also be individually reviewed as simulation doesn't make sense for most of them, elaboration does.

If you see any other tests that are similarly missed, please let me know (or fix them if you have time :wink: )

PR #1784, opened 2021-09-27
Drop yosys-sv

tgorochowik:
All the relevant patches are merged to mainline yosys and this fork is
not maintained anymore.


mithro:
@tgorochowik -- Really, this should become an empty package which just depends on the upstream Yosys now.

PR #1785, opened 2021-09-29
CI: conda: use only channels that were used to built package

kamilrakoczy:


tgorochowik (approved):

tgorochowik:
Verilator issues seem to be a problem with conda packages. Let's solve this separately (gcc should be required by the conda package).

PR #1786, opened 2021-09-29
Random collection of test fixes

MikePopoloski:
Blacklisting one ivtest:
- event_array is marked as "should fail" but actually is totally fine (and passes in all major simulators)

Blacklisting three yosys tests:
- load_and_derive and ondemand appear to be helper files part of a larger test and do not pass in isolation (and also have some invalid syntax naming a module "ref" which is a keyword)
- mem2reg_bounds_tern references an out of bounds array element at compile time -- you could argue this should pass with a warning but it's a nonsensical operation

Two assert uvm tests were invalid; they had a begin/end pair as the action block of a deferred assertion, which is disallowed by the LRM:
```
../..tests/chapter-16/16.2--assert0-uvm.sv:60:47: error: deferred assertion action must be a subroutine call
            assert #0 (m_if.a != m_if.b) else `uvm_error(label, $sformatf("assert failed :assert: (False)"));
                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../home/runner/work/sv-tests/sv-tests/third_party/tests/uvm/src/macros/uvm_message_defines.svh:153:4: note: expanded from macro 'uvm_error'
   begin \
   ^~~~~~~
```

Also tell the slang runner to only do parsing for the two black-parrot tests that cannot be elaborated due to invalid parameter values / top module listings.

### DIFF
--- a/conf/generators/meta-path/yosys-simple.json
+++ b/conf/generators/meta-path/yosys-simple.json
@@ -5,5 +5,5 @@
 		["tools", "yosys", "tests", "simple"]
 	],
 	"matches": ["*.v"],
-	"blacklist": ["hierdefparam.v", "memory.v", "values.v", "func_width_scope.v", "named_genblk.v"]
+	"blacklist": ["hierdefparam.v", "memory.v", "values.v", "func_width_scope.v", "named_genblk.v", "mem2reg_bounds_tern.v"]
 }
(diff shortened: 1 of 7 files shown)


tgorochowik (approved):
Thanks!

PR #1794, opened 2021-10-05
Use ubuntu-latest in CI everywhere

tgorochowik:
This is to make it consistent in all jobs. This should also fix moore
and verible.


mithro (approved):
LGTM.

hzeller (approved):
LGTM
That would explain the glibc issue.

PR #1796, opened 2021-10-06
More recategorization of tests after addition of elaboration mode

MikePopoloski:
This adds the elaboration mode to a bunch of tests that were missed when the big diff was landed (notably ivtests can elaborate just fine, and in addition most (all?) of the should_fail ivtests are things that should be caught at elaboration time so we shouldn't lose those tests).

Additionally I removed elaboration from a few tests that are truly sim-only because they require some kind of failure or value checking at runtime.


MikePopoloski:
That's certainly more complete but perhaps not worth the effort. Either way, outside the scope of this PR.

PR #1818, opened 2021-10-20
Add CI jobs to test makefiles

tgorochowik:


tgorochowik:
CC @hzeller the issue with Verible install can be seen here

PR #1859, opened 2021-11-16
Generate unique test name in path_generator

kamilrakoczy:
Currently all test cases from ``projf-explore/hello/hello-nexys/{A..I}/top.sv``are generated as: ``tests/generated/projf-explore/projf_top.sv``. This PR adds path from the test directory (with ``/`` changed to ``_``) to the name of the test and fixes this problem.

Now all test have unique test name.

### DIFF
--- a/generators/path_generator
+++ b/generators/path_generator
@@ -71,8 +71,9 @@ for cfg in glob.glob(os.path.join(conf_dir, 'generators', 'meta-path',
 
     for path in paths:
         for match in matches:
-            for f in glob.glob(os.path.abspath(os.path.join(third_party_dir,
-                                                            *path, match))):
+            search_path = os.path.abspath(
+                os.path.join(third_party_dir, *path, match))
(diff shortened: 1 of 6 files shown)


generators/path_generator
```diff
@@ -80,8 +80,19 @@ for cfg in glob.glob(os.path.join(conf_dir, 'generators', 'meta-path',
...
+                # check if this file is in folder inside test directory
+                folder_path = os.path.dirname(
+                    f.replace(
+                        os.path.commonprefix(
+                            [
+                                f,
+                                os.path.abspath(
+                                    os.path.join(
+                                        third_party_dir, *path, match))
+                            ]), "")).replace("/", "_")
```
tgorochowik, line 92: wow, nice formatting. I see that the checker agrees with this, but maybe we could do something about it?

tgorochowik (approved):

PR #1872, opened 2021-11-29
Add vanilla yosys with UHDM plugin

RRozak:


tools/runners/VanillaYosysPlugin.py
```diff
@@ -0,0 +1,76 @@
...
+#
+# SPDX-License-Identifier: ISC
+
+import os
+import sys
+
+from BaseRunner import BaseRunner
+
+
+class VanillaYosysPlugin(BaseRunner):
```
tgorochowik, line 18: We need to make sure it is clear that it's the UHDM plugin (as there are many different yosys plugins in general).

.github/workflows/sv-tests-ci.yml
```diff
@@ -39,11 +39,16 @@ jobs:
             deps: autoconf autotools-dev bison default-jre flex libfl-dev cmake pkg-config tclsh uuid-dev
             runners_filter: UhdmVerilator
           - name: yosys
-            deps: bison clang tcl-dev flex libfl-dev
+            deps: bison clang tcl-dev flex libfl-dev pkg-config libreadline-dev
+            runners_filter: Yosys
```
kamilrakoczy, line 43: With changed name of ``vanilla-yosys-uhdm-plugin`` binary, this is no longer needed.

kamilrakoczy (changes requested):
Please squash commits and make PR in ``yosys-uhdm-plugin-integration`` repository with all required changes.

.github/workflows/sv-tests-ci.yml
RRozak: Removed

kamilrakoczy (approved):

PR #1899, opened 2021-12-10
Move dependabot job to main CI and run after Summary

tgorochowik:


.github/workflows/sv-tests-ci.yml
```diff
@@ -218,3 +218,22 @@ jobs:
         uses: geekyeggo/delete-artifact@v1
         with:
           name: ${{ steps.get-artifacts-to-delete.outputs.artifacts }}
+  Automerge:
```
kamilrakoczy, line 221: If I'm not missing anything, this will also run on push to ``master`` branch or on schedule. I'm not sure what ``github.actior`` is set then, but it will probably be better to limit it only to PR.

kamilrakoczy (changes requested):

.github/workflows/sv-tests-ci.yml
tgorochowik: Good point, I'm not sure either. Let's add an explicit check

kamilrakoczy (approved):

PR #1905, opened 2021-12-11
Disable tests that do not have appropriate submodules defined.

wsnyder:
This updates the hdlconvertor blacklists to exclude some tests that are not legal and give simulation errors because submodules are not properly included/referenced, or they are not otherwise clean.

For specific messages on what modules are missing, look at the dashboard Verilator results for the tests.


tgorochowik (approved):
Thanks!

PR #1937, opened 2021-12-20
Blacklist bsg_ tests

wsnyder:
Blacklists some bsg_ tests that have syntax errors.


tgorochowik (approved):

mithro:
@taylor-bsg

taylor-bsg:
@tgorochowik Is there an easy way to get a link to the prior sv-tests report so I can see what errors were being reported for these tests before they were blacklisted? 

As an aside, it seems to me like there should be some kind of description of the problem on a pull request that black lists tests. And my github ID is not a secret, you can add me to the reviewer list. =)

wsnyder:
@taylor-bsg 

>bsg_cache_to_dram_ctrl.v

%Error: /svaha/wsnyder/SandBox/sv-tests/third_party/cores/basejump_stl/bsg_cache/bsg_cache_to_dram_ctrl.v:10:10: Cannot find include file: bsg_dmc.vh

>bsg_mem_banked_crossbar.v

%Error: /svaha/wsnyder/SandBox/sv-tests/third_party/cores/basejump_stl/bsg_mem/bsg_mem_banked_crossbar.v:126:19: syntax error, unexpected '/', expecting ';'
(The macro is trying to make a module name with slash in it.)

>bsg_sparse_to_dense_boolean.v

%Error: /svaha/wsnyder/SandBox/sv-tests/third_party/cores/basejump_stl/bsg_misc/bsg_sparse_to_dense_boolean.v:51:9: syntax error, unexpected '.', expecting ')'
(Comma missing after each pin.)

>bsg_nonsynth_fsb_node_trace_replay.v

This wasn't blacklisted, but since we're having this discussion I would note that this concat into a lvalue into a system call

                       result_code = $sscanf(read_line, "%b\n", {op_r, data_r});

isn't supported by Verilator, and since bsg isn't a test library but rather a functional library, you might consider changing this line to load into a temporary instead of using a concat.  It's of course a reasonable ding against Verilator's IEEE support, so this test remains on.

PR #1941, opened 2021-12-22
Report: Reimplement test preview panel, reduce index.html file size

mglb:
* Reduced index.html file size

  `index.html` size changed from 9.4M to 385K.
  Details about every tested (tag, tool) pair (a "result cell") has been moved to 4490 (= number of result cells) config files with cumulative size of about 3.3M.
  Config file for a result cell is loaded only when its details panel is opened.

  Numbers above come from a report containing results for 14 tools and 326 tags.

  In case this amount of config files is problematic, it shouldn't be a problem to e.g. merge all configs of a single tag, reducing the number of files to 326.

* URL hash replaced with URL query parameters

  Before: `.../#moore|22.11|22.11--pragma-complex`
  After: `.../?v=moore+22.11+22.11--pragma-complex`

  The hash is still checked and parsed when `v` parameter is absent, so all existing links will work.

  I'm working on an option of generating a report with results split into multiple tables. It could be nice to use the hash as a table anchor.

* Support for browser history. The page title contains a name of currently selected cell result.

  The state is now correctly restored during history navigation. Creation of repeated consecutive history entries on every click (caused by iframes) has been mostly fixed.
  The title looks better in a browser history, example:

  ```
  surelog/uvm-req/22.7--timescale-basic-2 - SystemVerilog Report
  surelog/uvm-req/22.7--timescale-basic-2 - SystemVerilog Report
  slang/uvm-req/22.7--timescale-basic-2 - SystemVerilog Report
  odin/uvm-req/22.7--timescale-basic-2 - SystemVerilog Report
  odin/5.6.4/define-directive - SystemVerilog Report
  moore_parse/5.6.4/define-directive - SystemVerilog Report
  moore_parse/5.6.4/number_test_5 - SystemVerilog Report
  moore_parse/5.6.4/number_test_59 - SystemVerilog Report
  moore_parse/5.6.4/number_test_32 - SystemVerilog Report
  moore_parse/5.6.4/begin-keywords - SystemVerilog Report
  moore/5.6.4/begin-keywords - SystemVerilog Report
  SystemVerilog Report
  ```

* Refactoring mostly in `tools/sv-report` and report.js preparing for "multi-table report" feature and improving the code a bit.

### DIFF
--- a/tools/sv-report
+++ b/tools/sv-report
@@ -28,6 +28,7 @@ import urllib.parse
 import dataclasses
 import enum
 from typing import Dict, Any, List, Set, Iterable
+import json
 
 parser = argparse.ArgumentParser()
 
@@ -428,6 +429,31 @@ def renderInputFileToHTMLFile(
                 csspath=csspath, filename=input_file_path, code=code))
 
 
+def renderTagResultsConfig(
+        js_path: str, tool: str, tag: str, test_results: Iterable[TestResult]):
+    tool = tool.lower()
+    tag = tag.lower()
+    js_global_variable_name = f"config_loader_data['{tool}/{tag}']"
+
+    # Order of values in each entry:
+    # [name, status, log_html_path, first_input_file_html_path]
+    config = []
+    for test_result in test_results:
+        name = test_result.name
+        status = int(test_result.status == TestStatus.PASSED)
+        log_html_file = test_result.log_html_file
+        first_input_html = test_result.input_files[0] + ".html"
+        first_input_html = urllib.parse.quote(first_input_html)
+        config.append([name, status, log_html_file, first_input_html])
+
+    js_data = json.dumps(config, separators=(",", ":"))
+    code = f"{js_global_variable_name} = {js_data}"
+
+    os.makedirs(os.path.dirname(js_path), exist_ok=True)
+    with open(js_path, "w") as f:
+        f.write(code)
+
+
 def collect_logs(runner_name: str):
     tool_results = ToolResults()
 
@@ -435,6 +461,8 @@ def collect_logs(runner_name: str):
     # with size > minimum_throughput_file_size
     passed_tests_time = 0.0
     passed_tests_input_files_size = 0.0
+
+    rendered_count = 0
 
     @dataclasses.dataclass
     class TagStatuses:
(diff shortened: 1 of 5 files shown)


tools/sv-report
```diff
@@ -272,177 +217,430 @@ meta_tags, meta_urls = readConfig(args.meta_tags)
...
+# Data passed to report template
+report_data = ReportResults()
+
+PATH_WITH_LINE_NO_MATCHER = re.compile(
+    r"([^\s:\"'`/]*/[^\s:\"'`]+\.\w+)(:(\d+)(?::\d+)?)?")
+
 
-for tag in database:
-    tag_usage[tag] = False
+def convertPathsToRelativeLinks(text: str, relative_to: str):
```
tgorochowik, line 318: It's not clear what it does, please explain, including the regex pattern

tools/sv-report
```diff
@@ -272,177 +217,430 @@ meta_tags, meta_urls = readConfig(args.meta_tags)
...
+    tool = tool.lower()
+    tag = tag.lower()
+    js_global_variable_name = f"config_loader_data['{tool}/{tag}']"
+
+    # Order of values in each entry:
+    # [name, status, log_html_path, first_input_file_html_path]
+    config = []
+    for test_result in test_results:
+        name = test_result.name
+        status = 1 if test_result.status == TestStatus.PASSED else 0
```
tgorochowik, line 416: `status = int(test_result.status == TestStatus.PASSED)` if not, then I think it'd be better to expand this if

tools/sv-report
```diff
@@ -272,177 +217,430 @@ meta_tags, meta_urls = readConfig(args.meta_tags)
...
+        # Render the config
+        renderTagResultsConfig(config_js, runner_name, tag, tag_results.tests)
 
-        for tag in tags:
-            passed_count = tags[tag]["status"].count("test-passed")
-            tags[tag]["passed-num"] = passed_count
+    try:
+        with open(os.path.join(args.logs, runner_name, "version")) as f:
+            tool_results.version = f.read().strip()
+    except:
```
tgorochowik, line 617: Please don't leave bare except. At the very least use `except Exception`, a bare except will also catch things like ctrl+c etc which we don't want

tgorochowik: (here and in other places)

tools/sv-report
mglb: Documented.

tools/sv-report
mglb: Fixed.

tools/sv-report
mglb: Fixed

tgorochowik (approved):

PR #1959, opened 2021-12-23
Add results_group property (...and a new filter)

mglb:
The PR implements `results_group` property that can be added to tests. Separate table in the report is generated for each group.
For simplicity, the results group ID can only contain characters `[0-1a-z_]`. To assign nice display name to a group, add an entry to `RESULTS_GROUP_DISPLAY_NAMES` in `tools/sv-report`. By default no name is displayed.
The default group ID (the one assigned to tests without group set) is empty string `""`.

Patch that tags some tests and adds display names for a few groups: https://github.com/antmicro/sv-tests/commit/0bfdf01e4c7d3ecc11a28f6122d6e1d10b5463ac.patch

There are also a few other changes (sorry for "all in one" PR :) ):
* Slightly improved test details panel style
* Faster and better looking sorting. Fractions with the same value are additionally sorted, so that 0/42 < 0/1 < ... 2/6 < 1/3 < ... < 2/3 < 4/6 < ... < 1/1 < 42/42.
* Reimplemented filter. It would be pretty hard to add support for multiple tables to the previous one.

Screenshot 2021-12-27 at 07-09-53 SystemVerilog Report

---

Screenshot 2021-12-27 at 07-20-46 SystemVerilog Report

---

Screenshot 2021-12-27 at 07-24-08 moore_parse fx68k fx68k - SystemVerilog Report

### DIFF
--- a/conf/report/code-template.html
+++ b/conf/report/code-template.html
@@ -2,10 +2,10 @@
 <html>
   <head>
     <meta charset="UTF-8" />
-    <link rel="stylesheet" type="text/css" href="{{ csspath }}">
+    <link rel="stylesheet" type="text/css" href="{{csspath}}">
   </head>
   <body>
-    {{ filename }}
+    <header><h2>{{filename|e}}</h2></header>
     {{ code }}
   </body>
 </html>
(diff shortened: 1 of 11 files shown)


tgorochowik (approved):
LGTM, let's merge this and do the actual separation in new PRs.

PR #2085, opened 2022-01-18
Build moore from a submodule

tgorochowik:
Currently we always try to install latest moore. This way we cannot really track/control the version. Current version breaks the CI at the moment.


tgorochowik:
Needs: https://github.com/SymbiFlow/actions/pull/37

PR #2116, opened 2022-01-24
Update earlgrey configuration

RRozak:


conf/fusesoc-configs/earlgrey.yml
```diff
@@ -9,11 +9,11 @@
 
 name: earlgrey
```
kamilrakoczy, line 10: I think we should split this file to at least 2 targets - ``synth`` and ``sim``, in similar to ibex:
- https://github.com/chipsalliance/sv-tests/blob/master/conf/fusesoc-configs/ibex-synth.yml
- https://github.com/chipsalliance/sv-tests/blob/master/conf/fusesoc-configs/ibex-sim.yml

In synthesis we should use ``lowrisc:systems:chip_earlgrey_nexysvideo`` target and current one: ``lowrisc:dv:chip_verilator_sim`` for simulation.

kamilrakoczy (changes requested):

conf/fusesoc-configs/earlgrey.yml
RRozak: Done

conf/fusesoc-configs/earlgrey-sim.yml
```diff
@@ -9,12 +9,12 @@
...
-conf_file: build/lowrisc_systems_top_earlgrey_0.1/sim-icarus/lowrisc_systems_top_earlgrey_0.1.scr
-test_file: earlgrey.sv
+command: fusesoc --cores-root third_party/cores/opentitan run --flag=fileset_top --target=sim lowrisc:dv:chip_verilator_sim
+conf_file: build/lowrisc_dv_chip_verilator_sim_0.1/sim-verilator/lowrisc_dv_chip_verilator_sim_0.1.vc
+test_file: earlgrey-sim.sv
 timeout: 360
-compatible-runners: all
-type: parsing
+compatible-runners: verilator-uhdm verilator slang
+type: parsing elaboration
```
kamilrakoczy, line 20: This type should also include ``simulation``

kamilrakoczy (changes requested):

conf/fusesoc-configs/earlgrey-sim.yml
RRozak: done

kamilrakoczy (approved):

tgorochowik (approved):

PR #2136, opened 2022-01-26
Add 'elaboration' mode back to ivtests

MikePopoloski:
This was recently changed from elaboration to simulation but that's not really correct; the tests don't really require simulation, but they do need elaboration in order for a bunch of them to pass -- most of the ones that are marked as should_fail require semantic analysis, not just parsing. Without setting elaboration, tools will fail to correctly fail on those tests.

Signed-off-by: MikePopoloski <mike@popoloski.com>

### DIFF
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -21,7 +21,7 @@ templ = """/*
 :incdirs: {3}
 :tags: ivtest
 :results_group: imported
-:type: simulation parsing
+:type: simulation elaboration parsing
 {2}
 {4}
 */


MikePopoloski:
Bump on this!

tgorochowik:
Hi @MikePopoloski,

Thanks for this PR. However icarus is a simulator, therefore these tests now run in simulation mode.

The `shoud_fail` tests are problematic indeed, if nothing else proves to be a good solution then maybe we'll exclude the 'parsing' mode from them.

I agree that some of these tests should be good enough for elaboration only but I'd like to take a closer look at this before merging this (so we avoid changing this back and forth).

Sorry for the delay, we'll certainly revisit this to see what's best here

MikePopoloski:
Maybe I'm missing the nuance here, but I'm not sure why it's a problem that Icarus is a simulator? The tests are still marked as 'simulation' as well, so any simulator will still continue to run the same way it does before this diff. All this diff does is make sure that, if a tool supports elaboration, it will try to do elaboration on ivtests (which is required to successfully pass the should_fail tests). If a tool doesn't support elaboration its results will be unchanged.

This doesn't fix things for tools that are only parsers, as they will still fail to pass the should_fail tests, but that's true both before and after this diff, so it's kind of orthogonal to this change. Probably the correct thing to do for those tools is figure out how to apply different modes for different subsections of ivtests, but that's out of scope from this change and not something I'm personally interested in (as slang can perform elaboration).

MikePopoloski:
@tgorochowik could you explain more about what the controversy is here? Again, just to summarize my understanding of the changes after this patch:
simulators => no change
parsers => no changes
elaboration tools => will actually perform elaboration now, which is required for 100+ tests to pass

tgorochowik:
@MikePopoloski I agree with you that your patch improves handling of those tests. All I am trying to say is that it is not necessarily complete.

As this is a test suite dedicated for simulation, I expect that at least some of the "should fail" test fail during simulation (not in elaboration).

We have a developer going through these right now, sorry for the delay once again, and thank you for your contribution!

tgorochowik (approved):

MikePopoloski:
Thanks, I agree that this is not fully sufficient as some tests do require full simulation to pass, but there are only a handful and this does fix ~150 tests so it seems like an easy win.

MikePopoloski:
Just to follow up on this, I looked at the remaining ivtest failures for slang and none of them require simulation, so I think it's safe to say that simulation is not required for any of the ivtests.

PR #2139, opened 2022-01-26
Split swerv config into swerv-lint and swerv-synth

RRozak:


tgorochowik (approved):

RRozak:
swerv-lint should be changes to swerv-sim when the https://github.com/chipsalliance/Cores-SweRV/issues/108 will be solved.

PR #2179, opened 2022-02-01
Move RUNNER_PARAM assignment outside the define

RRozak:
Before this change, the `RUNNER_PARAM` variable in line 71 was always empty.


tgorochowik (approved):
Thanks!

PR #2315, opened 2022-02-23
Checkout only tests and cores in summary

kamilrakoczy:
This PR speeds up summary job by checking out only tests and cores instead of every submodule (including tools).


kamilrakoczy:
Time of summary job went down from 46m 42s from last scheduled job to 41m 5s in this PR.

tgorochowik (approved):

PR #2586, opened 2022-04-14
Remove VanillaYosysUhdmPlugin runner

kamilrakoczy:
With updated ``yosys`` version ``VanillaYosysUhdmPlugin`` and ``UhdmYosys`` runners are the same.
All required patches are already upstream in ``yosys``.

This PR removes ``VanillaYosysUhdmPlugin`` runner and adapts ``UhdmYosys`` to use ``systemverilog`` plugin (renamed ``uhdm``).

This PR needs to be merged after: https://github.com/antmicro/yosys-uhdm-plugin-integration/pull/640.


tools/runners/UhdmYosys.py
```diff
@@ -32,7 +32,7 @@ def prepare_run_cb(self, tmp_dir, params):
 
         # generate yosys script
         with open(yosys_scr, "w") as f:
-            f.write("plugin -i uhdm\n")
+            f.write("plugin -i systemverilog\n")
             f.write("read_uhdm slpp_all/surelog.uhdm\n")
```
tgorochowik, line 36: Would there be any benefit in also testing `read_systemverilog` directly? Or maybe doing this instead of `read_uhdm`?

tgorochowik (approved):

tools/runners/UhdmYosys.py
kamilrakoczy: I think we should use``read_systemverilog`` directly, as it is probably easier command for users. I've adapted ``UhdmYosys`` runner to use it.

kamilrakoczy:
This PR is ready for review. I've checked new failing tests and they should fail (we doesn't support features they are testing).

tgorochowik (approved):

PR #3176, opened 2022-08-28
Update README.md

srochel:
Added missing dependencies.


hzeller:
Is there a reason why this is still a draft ? (while being draft, it can't be merged)

PR #3208, opened 2022-09-03
Update slang runner flags to allow tests to pass even if they access out of bounds array elements

MikePopoloski:


MikePopoloski:
@tgorochowik can this be merged?

MikePopoloski:
LRM allows this to be a compile-time error or to just produce x's and continue on. VCS will warn and continue compiling so I think it's ok to leave the tests as-is.

PR #3343, opened 2022-10-05
Test Verilator with `--timing`

kbieganski:
Signed-off-by: Krzysztof Bieganski <kbieganski@antmicro.com>


kbieganski:
Requires https://github.com/verilator/verilator/pull/3657.

Also note that the old version of `vmain.cpp` needs to be provided for `UhdmVerilator` (until it's updated to Verilator 5).

kbieganski:
There are two actual regressions in Verilator 5:
* https://github.com/verilator/verilator/pull/3658 (already merged),
* https://github.com/verilator/verilator/pull/3659 (really an older bug, but made visible with changes in V5).

The remaining newly failing tests were false positives before V5. Now they are failing, as they should.

tgorochowik (approved):

PR #3447, opened 2022-10-22
Fix class_test_52.sv to be legal

wsnyder:
IEEE 11.4.12.1 disallows zero size replication in the removed line.


tgorochowik (approved):
Thanks!

Issue #2749, opened 2022-05-23
tools/sv-report very slow


hzeller:
It takes about 40 minutes CPU time for the sv-report tool to run and creating a report (And it runs in a single thread).
This is painfully slow.
Are there some simple things that can be done to speed it up ?

tgorochowik:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.051    0.051   63.582   63.582 sv-report:1(<module>)
      142    0.000    0.000    0.000    0.000 sv-report:120(escapeXmlAttribute)
        2    0.000    0.000    0.000    0.000 sv-report:124(escapeJsString)
       86    0.000    0.000    0.000    0.000 sv-report:131(surroundWithQuotes)
     3289    0.002    0.000    0.010    0.000 sv-report:136(numericSortKey)
     2275    0.001    0.000    0.002    0.000 sv-report:138(<lambda>)
       64    0.000    0.000    0.000    0.000 sv-report:142(numericSort)
        5    0.000    0.000    0.000    0.000 sv-report:146(numericDictSort)
       11    0.000    0.000    0.000    0.000 sv-report:147(<lambda>)
        3    0.000    0.000    0.000    0.000 sv-report:152(tagDictSort)
       42    0.000    0.000    0.000    0.000 sv-report:153(tagSortKey)
       42    0.000    0.000    0.000    0.000 sv-report:159(<lambda>)
     8741    0.008    0.000    0.076    0.000 sv-report:173(exists_and_is_newer_than)
     3009    0.005    0.000    0.033    0.000 sv-report:185(totalSize)
        2    0.001    0.000    0.002    0.001 sv-report:200(readConfig)
        1    0.000    0.000    0.000    0.000 sv-report:235(TestStatus)
     3051    0.002    0.000    0.004    0.000 sv-report:242(__str__)
        1    0.000    0.000    0.000    0.000 sv-report:246(TestResult)
        1    0.000    0.000    0.000    0.000 sv-report:272(TagInfo)
        1    0.000    0.000    0.000    0.000 sv-report:278(ToolInfo)
        1    0.000    0.000    0.000    0.000 sv-report:288(TagToolResult)
      280    0.000    0.000    0.000    0.000 sv-report:295(total_tests)
      126    0.000    0.000    0.000    0.000 sv-report:299(status)
        1    0.000    0.000    0.000    0.000 sv-report:310(ToolSummary)
[...]

PR #3498, opened 2022-11-03
Improve Summary step speed

tgorochowik:
Marking as draft for now to see how it influences the report generation speed.

Ref #2749

### DIFF
--- a/tools/sv-report
+++ b/tools/sv-report
@@ -820,13 +820,28 @@ for tool_name, tool_results in zip(runner_names, results):
 
 # Render input files
 rendered_count = 0
-for input_file in all_input_files:
-    input_file_html = os.path.join(out_dir, input_file + ".html")
-    if exists_and_is_newer_than(input_file_html,
-                                [input_file, args.code_template, __file__]):
-        continue
-    rendered_count += 1
-    renderInputFileToHTMLFile(out_dir, input_file_html, input_file)
+
+
+def render_batch(input_files):
+    batch_rendered_count = 0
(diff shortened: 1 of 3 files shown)


tgorochowik:
Summary step time is down to 14m (from 30m), although the biggest benefit seems to come from removing json/verilog output from the UhdmYosys runner

tgorochowik:
The `sv-report` script itself runs for 2 minutes (down from 19m e.g. here: https://github.com/chipsalliance/sv-tests/actions/runs/3381547846/jobs/5616179090).

So I would say the improvement is quite big.

The remaining time is downloading all the sources/generating the test cases (maybe the latter could be parallelized too?)

tools/sv-report
```diff
@@ -115,6 +115,8 @@ top_dir = os.path.abspath(os.curdir)
 logs_out_dir = os.path.join(out_dir, "logs")
 
 lex = lexers.get_lexer_by_name("systemverilog")
+formatter = HtmlFormatter(
```
kamilrakoczy, line 118: I think we can't use single ``formatter`` object, as it looks like ``HtmlFormatter`` changes its state when it formats file, e.g.: https://github.com/pygments/pygments/blob/3f4c7331a18b073209297a125131f860b4bd8220/pygments/formatters/html.py#L861

kamilrakoczy (changes requested):
1 small comment, other than that LGTM.

hzeller (approved):
Nice, a significant improvement!

PR #3572, opened 2022-11-19
Disable some illegal test cases

wsnyder:
This disables more illegal test cases.

iverilog

* writememb1: output directory doesn't exist
* writememb2: output directory doesn't exist
* writememh1: output directory doesn't exist
* writememh2: output directory doesn't exist
* writememerror: negative test
* uwire_fail: negative test

hdlconv:

* p180.sv: Illegal, Packet is not declared.
* p191.sv: Illegal, vector is redeclared.
* p706.sv: Illegal, cannot mix 2001 init with 1995 portlist.
* p710.sv: Illegal, cannot mix 2001 init with 1995 portlist.
* p773.sv: Illegal, T redeclared.
* p777.sv: Intentional error test per comments

utd-sv:

* utd-sv_chip8_config: Illegal, empty
* utd-sv_core1_config: Illegal, empty
* utd-sv_casez: Illegal, proc assign to wire.
* utd-sv_expression_tostring: Illegal, signals not defined
* utd-sv_forever-disable: Illegal, signals not defined
* utd-sv_fpu: Illegal, signals not defined
* utd-sv_fpu_div: Illegal, signals not defined
* utd-sv_module-instance: Illegal, signals not defined
* utd-sv_std-3.11.1-parameters: Illegal, signals not defined

projf:

* lib_null_ice40_projf_SB_IO: Illegal, no param values
* lib_null_ice40_projf_SB_PLL40_PAD: Illegal, no param values
* lib_null_xc7_projf_MMCME2_BASE: Illegal, no param values
* lib_null_xc7_projf_OBUFDS: Illegal, no param values
* lib_null_xc7_projf_OSERDESE2: Illegal, no param values

PR #3573, opened 2022-11-19
Disable tests that have missing module file references

wsnyder:
This disables tests that are referring to modules that are not found.  It's possible these could be fixed with appropriate includes or otherwise, but at present they are not valid.

Merge this after #3572 (might conflict if so I'll fix).


tgorochowik:
There are some conflicts indeed and I agree that these should probably work if we fix the configuration.

These are parse-only btw, so that's why some tools pass those anyway

wsnyder:
@tgorochowik thanks.

tgorochowik (approved):

PR #3574, opened 2022-11-19
Fix randsequence tests to self-test without UVM

wsnyder:
This converts some of the random tests to use the typical assert self-checking rather than requiring UVM.


tests/chapter-18/18.16--random-weighted-case-randcase_0.sv
```diff
@@ -21,3 +21,11 @@ function int F();
     endcase
     return a;
 endfunction
+
+module top;
+   int x;
+   initial begin
+      x = F();
+      $display(":assert: (10 == %d)", x);
```
tgorochowik, line 29: We need to add
```
:type: simulation elaboration parsing
```

`simulation` is the important bit - otherwise those asserts are not evaluated

tgorochowik (commented):
The changes you added to the tests look good (besides the lack of test type definition).

I don't think we should remove the UVM tests though, they were already given a separate category/tag so they don't pollute the chapters table - if you prefer, we can move them to a separate (uvm) directory, but I think we should keep them

wsnyder:
I'll update the types.

As to UVM, what is the additional coverage provided with having them?  If the base tests are self checking, and the UVM is well covered with existing similar tests in other areas, aren't these just replicating coverage to no effect.

Basically there's a cost to each failing test in that it needs to be analyzed manually, and each additional failure that overlaps existing coverage has a human cost.

But if you feel strongly I'll back out their removal.

tgorochowik:
I went through other uvm tests and we should indeed have the things you propose to remove covered. It should be safe to remove them here.

tgorochowik (approved):

PR #3673, opened 2022-12-13
runners: icarus: Fix top-level module specification

larsclausen:
Currently when specifying a top-level module for the icarus runner the `-s` switch is passed together with the top-level module name as a single parameter. Both are separated by a space, e.g. `-s top`. This results in the space being included in the top-level module name, resulting in errors such as

    error: Unable to find the root module " top" in the Verilog source.

Pass to the `-s` switch and the top-level module name as two separate parameters to fix this.

### DIFF
--- a/tools/runners/Icarus.py
+++ b/tools/runners/Icarus.py
@@ -27,7 +27,7 @@ class Icarus(BaseRunner):
         self.cmd += ["-o", ofile]
 
         if params['top_module'] != '':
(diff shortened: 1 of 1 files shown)


tgorochowik (approved):
LGTM, thanks for the fix. Although there seem to be no increase in tests passed so it's possible there are additional issues, any PRs improving the support of Icarus would be greatly appreciated!

PR #3712, opened 2022-12-20
Use older bazel

tgorochowik:
There is an issue with platform detection in bazel 6.0.0

Until this is fixed in Verible, use an older bazel version.


hzeller:
LGTM.

PR #3741, opened 2022-12-27
runners: icarus: Support all modes

larsclausen:
The icarus runner at the moment only supports parsing mode. Add support for the other modes as well.

 * For `preprocessing` the `-E` flag is added in which case only the preprocessor is being executed.
 * For `parsing` mode now the `null` backend is used to skip the code generation phase.
 * For `elaboration` nothing special has to be done since it is the default behavior
 * To support `simulation` a shell script is created that first executes the iverilog compiler and then if that is successful executes the resulting executable. This is similar to what other runners do.

### DIFF
--- a/tools/runners/Icarus.py
+++ b/tools/runners/Icarus.py
@@ -15,16 +15,25 @@ import os
 
 class Icarus(BaseRunner):
     def __init__(self):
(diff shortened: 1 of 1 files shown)


tgorochowik (approved):
Thank you for the improvements!

PR #3820, opened 2023-01-14
Fix wildcard equality operator tests with x/z in LHS

larsclausen:
For the wildcard equality operator only x and z bits of the right hand side value are considered as wildcard bits.

x and z bits on the left hand side are still compared and may result in the comparison itself returning x.

There are a few test entries for the wildcard equality operator test that do not consider this and expect the wrong value. Correct these tests to expect that the result is 1'bx.


larsclausen:
I think it says "may result" since the result can also be 1'b0 or 1'b1 if it is clear from other bits that the values are/are not equal. E.g. `4'bx101 == 4'b000x` will yield 1'b0.

PR #3923, opened 2023-02-04
Add help2man for Verilator and update Verilator

wsnyder:
This fixes the build error that was preventing dependabot from being able to upgrade Verilator.  It also does the Verilator update.


wsnyder:
@tgorochowik et al could you consider this? thanks.

tgorochowik (approved):

tgorochowik:
Sure, sorry for the delay

PR #3924, opened 2023-02-05
Disable ivtests that fail on other simulators

wsnyder:
New list of disables after review below:

Disable ivtests that fail on other simulators

* br_gh104a - Null object dereferenced
* br_gh230_iv - Some tools error others warn, it's not clear in LRM what the requirements are
* br_gh60a - Zero sized constant is illegal, but some tools warn and ignore
* br_ml20180227 - String type is not implicitly assignment compatible to a packed vector, but some allow
* comment1 - Negative test wrong, what is tested appears legal on other simulators and by IEEE
* concat4 - Checked values FAILED on multiple other simulators
* enum_test3 - Illegal, but some tooks just warn
* escaped_macro_name - Fails on others; does not appear to match IEEE's description of define identifiers
* func_init_var3 - Checked values FAILED on multiple other simulators
* pr1938138 - Negative test that in some tools throws a warning rather than fatal error
* pr2794144 - Unclear, most tools hard error but for some tools allow
* vlg_pr1587634_iv - Specified --top-module 'top' was not found in design (sv-test driver limitation)

Disable hdlconvertor tests that fail on other simulators

* "p504_2.sv" - MyBus not declared
* "p522.sv" - Slist not declared (vs SList)


wsnyder:
@tgorochowik et al could you consider this? thanks.

MikePopoloski:
Many of these appear to be very suspect. I didn't go through all of them but for example:

br961 - All three major simulators support this just fine. The point of declaration for the port is considered to be at the name itself, not prior to it during the data type.

br_gh127c - "assigning to input" is a little too simplistic. Unfortunately the LRM allows this for net ports, in the section on "Port coercion", and all three major simulators will follow suit. 

comment1 - The LRM says this in the section on comments:
> Block comments shall not be nested.

br_ml20181012d - You can't have an array of width zero, what would that even mean? LRM says this:
> Each fixed-size dimension shall be represented by an address range, such as [1:1024], or a single positive
number to specify the size of a fixed-size unpacked array, as in C.
This also fails on e.g. VCS.

MikePopoloski:
I would also say that it seems pretty valuable to have test coverage of cases that the LRM calls out as illegal, even if by default your tool treats it as a warning instead. Surely there's a way for you to promote warnings to errors such that you could still pass the test suite? Or if you don't have any warning at all for something the LRM says is illegal, that seems like a deficiency worth rectifying in your tool. Just throwing the tests away instead seems wrong.

wsnyder:
By nested comments isn't legal they mean you can't do `/* /* */  */` and the whole thing is commented out.  Please try the test on other simulators, I did.

As to "reg x[0]" while it might seem clear that is nonsensical, I suspect you'd agree this is also IEEE illegal under the same rule:

```
module test();
  parameter X = 0;
  reg illegal[X];
endmodule
```

After elaboration this reduces internally in my code to the same case the tests cover (reg illegal[0]).  While strange, I've seen this in the "real world" where "illegal" is never used (maybe X=0 disables some feature).  This is an error on only 1 of the big 3, and so I elected not to consider it a fatal error.   (FWIW C discovered years ago that having an array of zero size is useful, SystemVerilog hasn't caught up.)

I can't promote every warning to error, nor do I wish to disable some errors (like assigning to input which I agree is technically legal but almost always "wrong"), doing such would cause false failures or false passes.  For example there's a boatload of other lint problems all through the tests (which is expected and reasonable given the tests' purposes).  You are right I could promote certain warning enables just to match what this these tests expect, but realistically that doesn't help users in any way on real designs, so I'll likely instead just keep ignoring the test failures as I have been doing previously.  

Anyhow I'd greatly appreciate a review on what ones you think are reasonable to disable or not.

MikePopoloski:
Here are the ones I think are pretty uncontroversial:
* br_gh104a
* comment1 -- After looking at this again I agree with you, my interpretation of the rule was incorrect.
* concat4
* escaped_macro_name
* func_init_var3
* vlg_pr1587634_iv

Also the hdlconvert test disables seem fine. They make no attempt to elaborate correctly, and it's not strictly clear where the line is between just parsing and full elaboration (like does name resolution count or not). I'm curious though why you say the PLA tasks aren't part of IEEE -- there's a whole section on them in the LRM?

I can do another pass on the remaining ones in your list later today or tomorrow.

tgorochowik:
I am okay with merging the changes you guys agree on.

@MikePopoloski please let us know if you intend to do the another pass you mention - if not maybe we can selectively merge the ones that are not controversial and keep the rest for discussion?

MikePopoloski:
Apologies, I fell severely ill this past weekend and am still catching up with life. I'll come back to this soon.

MikePopoloski:
OK, here's the full list.

Uncontroversial (remove test)
------------------------
* br_gh104a
* comment1
* concat4
* escaped_macro_name
* func_init_var3
* vlg_pr1587634_iv

After some more thought (remove test)
-------------------------------------
* br_gh230_iv - some tools error on this but it's not exactly clear in the LRM what the requirements are, and some tools just warn, so it seems fine to remove this one
* br_gh60a - zero sized constant is clearly illegal (LRM says must be a positive number) but bizarrely tools have chosen to warn and ignore it so I guess we should remove this one too
* br_ml20180227 - illegal, string type is not implicitly assignment compatible to a packed vector, but some tools (VCS) allow it anyway so I guess we should remove
* enum_test3 - definitely illegal, LRM says "Any enumeration encoding value that is outside the representable range of the enum base type shall be an error" but some tools allow it anyway
* pr1938138 - illegal but some tools just warn instead
* pr2794144 - unclear, most tools hard error but for some reason VCS allows it

Legal (keep test)
------------
* br961 - not a circular reference
* br_gh127c - assigning to input nets is legal
* br_gh127f - same as above
* pr2913927_iv - Having too many vector digits is not illegal. LRM says "If the size of the unsigned number is larger than the size specified for the literal constant, the unsigned number shall be truncated from the left." All major tools support this.
* sv_class4_iv - same as above
* vlg_pr509_iv - Assigning a wire to itself may be pointless but it's not illegal. All major tools support this.

Illegal (keep test)
-------------------
* br_ml20181012d - not legal in SV, all major tools throw an error
* enum_test5 - LRM is clear, "If the integer value expression is a sized literal constant, it shall be an error if the size is different from the enum base type, even if the value is within the representable range." VCS and ModelSim both error on this.

Also would still like to hear about the PLA tasks that I asked about in my earlier message.

wsnyder:
Thanks for the review, I updated the pull request to match the ones you indicated.

You're right the PLA is still formally part of IEEE, so left those in.

I still think that allowing overflows of constants e.g. "2'd1234" is cruel to users as it's allowing something that can only be a mistake, but I'll leave the tests matching what the LRM says, and I will ask IEEE to allow tools to consider this illegal in the next LRM.

MikePopoloski:
To be clear, I think it's totally reasonable for a tool to default to an error for this case, but it seems like philosophically you want a way to disable such errors when the LRM says such a construct is legal, because surely there are large codebases out there that do this and perhaps even work correctly even though it's misleading and ugly, and it's fair for sv-tests to test that tools can be compliant with such behaviors.

Also the diff lgtm but I don't have merge privileges so you'll have to wait for @tgorochowik

tgorochowik:
Thank you both @wsnyder @MikePopoloski ! I appreciate the changes and the discussion

PR #4046, opened 2023-03-01
ci: update custom runners labels

AdamOlech:


tgorochowik:
The failures are unrelated to the change, and the change is important to avoid picking up the organization wide runners. Merging.

Issue #3776, opened 2023-01-05
ariane file set is out of sync with ariane core


alaindargelas:
Surelog and sv-lang have missing definitions. The project definition is not following what the ariane core build system requires anymore,

tgorochowik:
Fixed by #4111 and #4128

PR #4111, opened 2023-03-12
Fix Ariane file list generation

MikePopoloski:


MikePopoloski:
Resolves #3776

tgorochowik (approved):
Nice, this is a great improvement! Thank you!

I think we could try to move your code that parses the `.cva6` file to a new function and invoke it in the loop above when we encounter `-f` in the `verilate_command`. This would be more future-proof. But I think we can merge it as is and do the change separately.

PR #4128, opened 2023-03-15
Use a list for storing ariane source files

tgorochowik:
Those files need to be provided in the same order they are listed, the order cannot be arbitrary.


tgorochowik:
@MikePopoloski with this minor change, Ariane is green for slang

PR #4152, opened 2023-03-18
Update 6.23--type_op_compare.sv

Thomasb81:
Fix syntax error : extra ')'.

To be note :
- some tool does not notice the syntax error
- the syntax error is present in text example Ieee1685-2017 page 131
image


Thomasb81:
wrong analyze !

PR #4603, opened 2023-06-17
Remove some awkwardly placed redundant semicolons

AndreasLoow:
Noticed some random semicolons in some test files for chapter 5 that probably should not be there


tgorochowik (approved):
Looks good, thank you for the contribution!

The CI fail is unrelated and already fixed in the main branch, merging!

PR #4620, opened 2023-06-21
Fix moore build in CI

larsclausen:
Moore build is currently broken in CI since one of its dependencies requires rust 1.60, but the CI only install rust 1.59.

Bump the rust version for the CI to 1.60 to fix this.


tgorochowik:
Thank you!

tgorochowik (approved):

Issue #4549, opened 2023-06-07
tree-sitter-verilog: essentially all tests fail with "Incompatible Language version"


hzeller:
Looking at the outputs of the tree-sitter-verilog tests, (here an example)
https://chipsalliance.github.io/sv-tests-results/?v=tree_sitter_verilog+uvm+adding_constraints_to_scope_variables_1

... they all fail with some obscure error message:
`"Incompatible Language version 13. Must not be between 9 and 12"`

... I suspect this is some sort of installation problem of whatever underlying parser it uses ?

tgorochowik:
Fixed by updating base tree_sitter package in #4685.

PR #4685, opened 2023-07-04
Unfreeze tree sitter

tgorochowik:
The version we used was pretty old, there were lots of changes in tree sitter verilog which hopefully allow us to use the latest py tree sitter.

Ref: #4549


tgorochowik:
Seems to have helped.

PR #4765, opened 2023-07-23
Avoid race conditions between `assign` and `initial`

larsclausen:
The initial execution of a `assign` statement is scheduled in the same time slot as an `initial` block.

There are a few tests that use an `initial` block to evaluate whether an `assign` statement was correctly executed. But this leads to an inherit race condition as the `initial` block might run before the `assign` has been executed.

Switch these tests to use a `final` block instead to avoid the race condition.

### DIFF
--- a/conf/generators/templates/assignment-sim.json
+++ b/conf/generators/templates/assignment-sim.json
@@ -12,7 +12,7 @@
 		"reg [31:0] a;",
 		"wire [31:0] b;",
 		"assign b = 32'd{3};",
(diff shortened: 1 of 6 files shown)


larsclausen:
Hm, looks like Odin does not support final. But then again Odin does not support simulation and these are all simulation tests. Maybe just disable simulation tests for Odin?

tgorochowik (approved):
The change looks good and you suggestion sounds good, thanks

PR #4926, opened 2023-08-31
Rename SweRV to VeeR & add VeeR EL2

robertszczepanski:
SweRV is officially renamed to VeeR, this PR renames it in sv-tests. There are also new tests for VeeR EL2 core.


tmichalak:
LGTM

PR #4943, opened 2023-09-04
Remove VeeR asserts from simulation builds

### LINKED ISSUE #4941: Fix Veer-EL2 with Verilator
@robertszczepanski I appreciate adding Veer cores, since they are Chips Alliance cores it's great to track them.

Veer-el2 supports Verilator as a simulator, however under sv-tests it fails due to a "##" unsupported error, perhaps because RV_ASSERT_ON is misset. Can you please fix that?

Also in the future as described in #1871, it would be much appreciated if all simulator results are audited before adding new tests to avoid failures such as these.

robertszczepanski:
Fixes #4941.


tmichalak (approved):
LGTM

PR #4953, opened 2023-09-05
Refer to specific commits in GitHub links

kiryk:
This change adds commit hashes to project links that refer to GitHub repositories.

It should affect only the links used in the HTML reports of the results.


tmichalak (approved):
LGTM

PR #5525, opened 2024-01-12
Minor enhancements to the Yosys runner

daglem:
* For simulation tests, suppress display task output from elaboration
* More robust detection of top module by fallback to -auto-top
* Added 'clean' pass to avoid passing unused memory to memory_collect


tgorochowik (approved):
LGTM>, thanks

PR #5620, opened 2024-02-04
Try adding yosys-slang

povik:


povik:
Apologies, this was a test PR which I wanted to open against my own fork

PR #5808, opened 2024-07-08
Fix ariane generator (again)

### LINKED ISSUE #5807: Ariane core test is still broken
The recent change definitely helped things, but the build is still not configured correctly. Most relevant slang error:
```
../../../third_party/cores/ariane/core/cache_subsystem/cva6_hpdcache_subsystem.sv:171:12: error: 'hpdcache_typedef.svh': No such file or directory
  `include "hpdcache_typedef.svh"
           ^~~~~~~~~~~~~~~~~~~~~~
```

Moschn:
Ariane seems to require recursive filelists. This commit adds recursive filelist parsing to the Ariane generator.

I am trying to set up a local instance of sv-tests, but compiling all the runners takes a while. In the meantime, we can also check what the CI says.

This commit (hopefully) fixes #5807

### DIFF
--- a/generators/ariane
+++ b/generators/ariane
@@ -78,6 +78,9 @@ def process_filelist(filelist):
                 incdirs += l.partition("+incdir+")[2] + ' '
             elif l.startswith("+define+"):
                 defines += l.partition("+define+")[2] + ' '
+            elif l.startswith("-F"):
+                # recursively parse filelists
+                process_filelist(l.partition(' ')[2])
             elif l.endswith(".sv"):
                 sources.append(l)
 


tgorochowik:
Thanks!

Note that you should be able to run sv-tests with even a single runner (meaning you don't have to compile them all from scratch, moreover you can use tools from outside of this repo if that's easier). So in this case it should be safe to verify that with just slang (as this one is expected to work for Ariane)

tgorochowik:
```bash
$ cat new_passes.csv 
sv_parser,ariane
VeribleExtractor,ariane
Slang_parse,ariane
```

Slang_parse is good, let's merge it, thanks!

tgorochowik (approved):

PR #5866, opened 2024-08-06
[Tools] Add the new tool - circt-verilog

cepheus69:
The CIRCT project has started to adapt to the IEEE 2017 standard of SystemVerilog and introduced a new tool, circt-verilog. Our compilation flow is based on the slang frontend compiler. Initially, SystemVerilog code is compiled into a slang Abstract Syntax Tree (AST). We then traverse the AST to generate Moore Intermediate Representation (IR). Several transformation and optimization passes are performed at the Moore dialect layer to produce more efficient, concise, and performant CIRCT Core IR code. As our development progresses, more comprehensive and in-depth testing will be conducted to ensure the stability and reliability of the compilation flow and related tools.

This PR aims to add the new tool - `circt-verilog`. Here are some key points about the changes:
1. Add a new tool runner - `circt-verilog`
2. Append related configuration into the CI yaml
3. Support using makefile to compile `circt-verilog`

- [Slang](https://sv-lang.com)
- [Moore Dialect](https://circt.llvm.org/docs/Dialects/Moore)
- [circt-verilog tool](https://github.com/llvm/circt/blob/main/tools/circt-verilog/circt-verilog.cpp)


tgorochowik (commented):
Thanks for the PR, would you mind taking a look at the linter failures reported by the CI?

cepheus69:
> Thanks for the PR, would you mind taking a look at the linter failures reported by the CI?

Thanks so much for your review! 😸 Yes. I am trying to fix failed code-quality check. It reports that the runner's python code did not pass the git diff check. I'm a little unsure of what's causing this error, I've checked for uncommited changes and worktree both locally and on the fork branch.

tgorochowik:
The CI check runs a python formatter on the source code and then checks if the formatter produced any differences - it did in this case, that's why it failed.

After formatting the code correctly, the diff should be empty and the CI job will pass.

You can treat the diff you see in the CI logs as a patch that need to be applied to your code, alternatively you can run `make format` which should apply the correct formatting (assuming you have yapf installed)

cepheus69:
> The CI check runs a python formatter on the source code and then checks if the formatter produced any differences - it did in this case, that's why it failed.
> 
> After formatting the code correctly, the diff should be empty and the CI job will pass.
> 
> You can treat the diff you see in the CI logs as a patch that need to be applied to your code, alternatively you can run `make format` which should apply the correct formatting (assuming you have yapf installed)

I installed ruff to format it.

cepheus69:
It seems that the runners for testing CI flow had lost connection.

tgorochowik (approved):

tgorochowik:
circt-verilog is now part of the main report, congrats!

Is there anything we can do to speed up the CI fo this tool? In total, it takes over 40 minutes to finish, 7 of which is for cloning submodules (are all of them needed? Can we use shallow clones instead?), the build process takes 29 minutes (is everything that is built used? maybe some optional tools can be disabled here).

Tests themselves execute in under a minute.

This is not a super big problem, as we do have a tool which takes even a bit longer, but it would be nice if you could try optimizing the CI time a bit, thanks!

PR #5867, opened 2024-08-06
Add PR write permissions to the comment bot

tgorochowik:
This is required to post a comment after the latest changes in GH api.


tgorochowik:
The comment job runs from the context of the main branch, so this has to bee merged before it can be properly tested

PR #6076, opened 2024-09-16
Add CI job for tree-sitter-systemverilog

gmlarumbe:
The parser is missing in the public report since I forgot to add a CI job for it. 

I simply added an entry in the RUNNERS_TARGETS var of the runners.mk file. Not sure if something else is needed. I also added it to the list of supported tools in the README.md

Let me know if something else is needed.

Thanks!

### DIFF
--- a/.github/workflows/sv-tests-ci.yml
+++ b/.github/workflows/sv-tests-ci.yml
@@ -30,6 +30,8 @@ jobs:
           - name: sv-parser
             deps: cargo
             rust_ver: "1.74"
+          - name: tree-sitter-systemverilog
+            deps: gcc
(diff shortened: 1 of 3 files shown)


tgorochowik:
Thanks!
To make it actually run, it needs to be added in the CI configuration (github actions in this case)

Please look here for reference: https://github.com/chipsalliance/sv-tests/blob/master/.github/workflows/sv-tests-ci.yml#L16

tgorochowik:
I think it was set to an older version because something was failing to build, feel free to bump it to a newer release, if the CI is still green after the upgrade, we can merge it

tgorochowik (approved):

PR #6077, opened 2024-09-16
Bump tree-sitter version to 0.23.0

gmlarumbe:
Separate PR from #6076 to update tree-sitter to a newer version after adding tree-sitter-systemverilog CI jobs.

### DIFF
--- a/conf/requirements.txt
+++ b/conf/requirements.txt
@@ -1,6 +1,6 @@
 pyyaml
 jinja2
-tree_sitter==0.20.4
+tree_sitter==0.23.0
 pygments
 yapf==0.30.0
 psutil


gmlarumbe:
For some reason pip cannot find the latest tree-sitter version (https://pypi.org/project/tree-sitter/):
```
ERROR: Could not find a version that satisfies the requirement tree_sitter==0.23.0 (from versions: 0.0.1, 0.0.2, 0.0.3, 0.0.4, 0.0.5, 0.0.6, 0.0.7, 0.0.8, 0.0.9, 0.0.10, 0.1.0, 0.1.1, 0.2.0, 0.2.1, 0.2.2, 0.19.0, 0.20.0, 0.20.1, 0.20.2, 0.20.3, 0.20.4)
ERROR: No matching distribution found for tree_sitter==0.23.0
```
Might it have to do with python version?

On the other hand, tree-sitter-verilog seems unmaintained. There is an open PR from some months ago to update it to version 0.21: https://github.com/tree-sitter/tree-sitter-verilog/pull/76

tgorochowik:
@gmlarumbe sounds great to me! Please try to adjust the CI so it works the way you propose and we can merge it if green, thanks!

gmlarumbe:
@tgorochowik ,

All tests are failing for `tree-sitter-systemverilog` due to this version mismatch issue (see https://github.com/chipsalliance/sv-tests/pull/7024):
```
Incompatible Language version 15. Must be between 13 and 14

[tests/generic/member/class_member_test_0.sv](https://chipsalliance.github.io/sv-tests-results/tests/generic/member/class_member_test_0.sv.html)::: error: unknown error: Parsing failed
```
I tried updating the `conf/requirements.txt` but required `tree-sitter` version is not available with `pip` yet.

Do you have any suggestion about how to add this latest version as a dependency? Would this latest version also be used by `tree-sitter-verilog`?

Thanks!

tgorochowik:
@gmlarumbe what is the exact version that you need?

if the release on pypi is not sufficient, maybe we could switch to using git directly? i.e https://github.com/tree-sitter/py-tree-sitter/tree/v0.24.0 ?

this would be
```
git+https://github.com/tree-sitter/py-tree-sitter.git@v0.24.0
```
in the requirements file.

If that would help, we could also use a specific commit instead of the version tag (if you need something from the dev branch for example)

gmlarumbe:
Hi @tgorochowik ,

The required version would be at least v0.25.2. This version still does not have a release for the Python bindings that are used in the runner. 

In my opinion the simplest fix would be to change the runner to use the `tree-sitter-cli` instead of the Python bindings. This would require installation of `tree-sitter-cli` (https://github.com/tree-sitter/tree-sitter/tree/master/cli). There are pre-built binaries (https://github.com/tree-sitter/tree-sitter/releases/tag/v0.25.3).

What would be the way to include this pre-built binary of `tree-sitter-cli` in the `conf/requirements.txt` file?

Thanks again!

zoomer-k:
python binding is updating now
https://github.com/tree-sitter/py-tree-sitter/pull/333

not merged yet

tgorochowik:
@gmlarumbe We already use npm in the build process, I guess we could install `tree-sitter-cli` through npm before building (instead of using pip).

Alternatively, I think it should be okay to use pip with git pointing to the current last commit from https://github.com/tree-sitter/py-tree-sitter/tree/0.25 (so this is kind of what @zoomer-k suggest - thanks for this btw) - this should probably be simpler as there are fewer changes

tgorochowik:
I pushed a commit that should switch the package to a dev release, let's see if this helps. If not, feel free to go the `npm` way

gmlarumbe:
I guess this would be the simplest solution. However it seems to be incompatible with current runner of `tree-sitter-verilog`. It throws the following error:
```
09:27:37 | AttributeError: type object 'tree_sitter.Language' has no attribute 'build_library'
```
Due to the latest line of its runner in `runners.mk`:
```
$(INSTALL_DIR)/lib/tree-sitter-verilog.so:
	mkdir -p $(INSTALL_DIR)/lib
	cd $(RDIR)/tree-sitter-verilog && npm install
	/usr/bin/env python3 -c "from tree_sitter import Language; Language.build_library(\"$@\", [\"$(abspath $(RDIR)/tree-sitter-verilog)\"])"
```
Probably the runner of `tree-sitter-verilog` should be updated. Related GitHub issue: https://github.com/tree-sitter/py-tree-sitter/pull/145#issuecomment-1969482456

Any idea of what would be the best option?

Thanks again for taking the time to find a solution.

tgorochowik:
@gmlarumbe would you be willing to update `tree-sitter-verilog` to use the pre-compiled wheel as suggested in the linked issue?

Alternatively we could try to come up with a way to isolate the runners a bit more and install dedicated dependencies (i.e. separate requirements, maybe venvs?) but that could be a bit more work I imagine.

gmlarumbe:
I force-pushed some changes that fix the compilation of the shared object for `tree-sitter-verilog`. 

To fix the Code Quality Checks would it be possible to add Python 3.10 to the `conf.environment.yml` file and coexist with current Python 3.8?

tgorochowik:
do we even need python 3.8? would it be possible to just switch to 3.10 completely?

tgorochowik:
I am not sure if the conda environment is still even used though? feel free to modify it if it still is used

tgorochowik (approved):

gmlarumbe:
Hi @tgorochowik ,

I would say they are not yet since it seems almost all tests are failing both for `tree-sitter-verilog` and `tree-sitter-systemverilog`.

Would it be possible to check the SV test logs of this integration CI?

Thanks!

gmlarumbe:
Seems that newer versions of tree-sitter Python bindings recommend to download the language grammar from `pip` as a Python module:
```python
import tree_sitter_python as tspython
from tree_sitter import Language, Parser

PY_LANGUAGE = Language(tspython.language()) 

parser = Parser(PY_LANGUAGE)
```
However `tree-sitter-systemverilog` is not available with `pip` and `tree-sitter-verilog` builds are failing (https://www.piwheels.org/project/tree-sitter-verilog/).

Python modules to import for the runners are located at `third_party/tools/tree-sitter-systemverilog/bindings/python/tree_sitter_verilog`. Could it be possible to add this path to the `$PYTHONPATH` or something similar so that it can be imported for the runner? Otherwise the use of `tree-sitter-cli` seems the only option possible.

Thanks again!

tgorochowik:
> Could it be possible to add this path to the `$PYTHONPATH` or something similar so that it can be imported for the runner? Otherwise the use of `tree-sitter-cli` seems the only option possible.


Would `sys.path.append(..)` work?

gmlarumbe:
> Would sys.path.append(..) work?

I just tried but got the following error:

```
module 'tree_sitter_verilog' has no attribute 'language'
```

It seems the grammar parser bindings are not working. I regenerated them on `tree-sitter-systemverilog` but I am afraid this approach will not work for `tree-sitter-verilog`. 

At this point I would suggest using `tree-sitter-cli` for both applications rather than Python bindings.

tgorochowik:
> At this point I would suggest using `tree-sitter-cli` for both applications rather than Python bindings.

Sounds good

gmlarumbe:
Hi @tgorochowik ,

I have made the changes to use `tree-sitter-cli` but installation through `npm` fails:
```
Run export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
17:22:50 | /bin/sh: 1: file: not found
17:22:50 | npm install --prefix /root/sv-tests/sv-tests/out/runners tree-sitter-cli
17:22:51 | npm ERR! code 1
17:22:51 | npm ERR! path /root/sv-tests/sv-tests/out/runners/node_modules/tree-sitter-cli
17:22:51 | npm ERR! command failed
17:22:51 | npm ERR! command sh -c node install.js
17:22:51 | npm ERR! /root/sv-tests/sv-tests/out/runners/node_modules/tree-sitter-cli/install.js:43
17:22:51 | npm ERR! const arch = platform?.arch[process.arch];
17:22:51 | npm ERR!                       ^
17:22:51 | npm ERR! 
17:22:51 | npm ERR! SyntaxError: Unexpected token '.'
17:22:51 | npm ERR!     at wrapSafe (internal/modules/cjs/loader.js:915:16)
17:22:51 | npm ERR!     at Module._compile (internal/modules/cjs/loader.js:963:27)
17:22:51 | npm ERR!     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
17:22:51 | npm ERR!     at Module.load (internal/modules/cjs/loader.js:863:32)
17:22:51 | npm ERR!     at Function.Module._load (internal/modules/cjs/loader.js:708:14)
17:22:51 | npm ERR!     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
17:22:51 | npm ERR!     at internal/main/run_main_module.js:17:47
17:22:51 | 
17:22:51 | npm ERR! A complete log of this run can be found in:
17:22:51 | npm ERR!     /root/.npm/_logs/2025-03-20T17_22_50_528Z-debug-0.log
17:22:51 | make: *** [tools/runners.mk:83: /root/sv-tests/sv-tests/out/runners/node_modules/tree-sitter-cli/tree-sitter] Error 1
```

I can install it locally in my machine. Do you happen to know what could be wrong with `npm install`? 

Related thread: https://stackoverflow.com/questions/67399785/how-to-solve-npm-install-error-npm-err-code-1

Thanks!

tgorochowik:
The answer suggests, the latest npm version is not compatible with tree-sitter-cli.

Maybe we can try forcing a specific version with `npm install npm@x.x.x -g`? We can use the version that works for you locally.

Note that js and npm are not my native languages so there may be a better way to handle this

gmlarumbe:
I also tried downloading a precompiled binary via `wget` but still getting errors while generating the parser:
```
15:31:40 | const result = await import(getEnv("TREE_SITTER_GRAMMAR_PATH"));
15:31:40 |                ^^^^^
15:31:40 | SyntaxError: Unexpected reserved word
```

This error also seems to be related to an old version of Node: https://github.com/conventional-changelog/commitlint/issues/2256

Any way to get a newer Node version in the environment?

tgorochowik:
Maybe we could use https://github.com/actions/setup-node to select a specific version?

gmlarumbe:
That could solve the issue.

Would you mind updating the GitHub CI workflow in the PR branch? I am not sure what would be the proper place to do it.

Thanks!

tgorochowik:
@gmlarumbe could you please check if everything works as expected after the changes from @kbieganski ?

tools/runners/tree_sitter_systemverilog.py
```diff
@@ -10,94 +10,25 @@
...
-            except Exception as e:
-                self.log_error(src, '', '', 'unknown error: ' + str(e))
-                self.ret = 1
-        usage = resource.getrusage(resource.RUSAGE_SELF)
-        profiling_data = (usage.ru_utime, usage.ru_stime, usage.ru_maxrss)
-
-        return (self.log, self.ret) + profiling_data
+    def prepare_run_cb(self, tmp_dir, params):
+        # src_dir = os.path.join(
+        #     (os.environ['THIRD_PARTY_DIR']),
```
tgorochowik, line 28: The commented out code needs to go before merging

tools/runners.mk
```diff
@@ -72,22 +72,26 @@ $(INSTALL_DIR)/bin/zachjs-sv2v:
...
+
+tree-sitter-verilog: $(INSTALL_DIR)/bin/tree-sitter
+	(export PATH=$(INSTALL_DIR)/bin/:${PATH} && \
+		cd $(RDIR)/tree-sitter-verilog && tree-sitter generate)
+	# cd $(RDIR)/tree-sitter-verilog && tree-sitter generate
+
+$(INSTALL_DIR)/bin/tree-sitter:
+	wget https://github.com/tree-sitter/tree-sitter/releases/download/v0.25.3/tree-sitter-linux-x64.gz
+	gunzip tree-sitter-linux-x64.gz
+	# chmod +x tree-sitter-linux-x64
```
tgorochowik, line 91: same here

tgorochowik (changes requested):

gmlarumbe:
Hi @tgorochowik ,

Sorry for the delay in the answer. I have been a bit busy lately.

Regarding the error I do not really think pushed changes have to do anything with it. Let me know if I need to do something else. 

Thanks!

PR #6414, opened 2024-11-13
Disable bsg_mem_multiport_latch* and implicitport/2/3/6

wsnyder:
Disables more illegal tests:

* bsg_mem_multiport_latch_write_banked_bypassing.sv - Bad commenting makes illegal "end"
* bsg_mem_multiport_latch_write_banked_bypassing_sync.sv - Likewise
* implicitport2 - Already in disables, but had extra _ so was not blacklisted.  (Test has multiple modules out of order, which breaks test harness.)
* implicitport3 - Likewise.
* implicitport6 - Likewise.


tgorochowik (approved):
Thanks

PR #6990, opened 2025-02-27
Update dependencies

majanicki:
One of the pip packages requires newer `setuptools` version, which causes the CI to fail. Updating `setuptools` before installing packages should solve the issue. 

Also Ubuntu24 does not support python 3.7 so I updated it to 3.8 (newer versions do not have `lib2to3`)


kbieganski (approved):
LGTM

PR #7081, opened 2025-04-04
Bump Verible to `7aae5c08`

kbieganski:
Verible installation is [done differently](https://github.com/chipsalliance/verible/pull/2298) now, reflected by this change.


hzeller:
that failure and error message look somewhat strange. What version of bazel is actually used ? (Verible should work with 6 up to 8, but not <= 5 anymore).

PR #7133, opened 2025-04-30
Update yosys-slang flags

povik:
Transfers more flags from the slang runner to yosys-slang to work around quirks of particular test cases. Add additional flags to make yosys-slang more lenient.

### DIFF
--- a/tools/runners/yosys_slang.py
+++ b/tools/runners/yosys_slang.py
@@ -42,7 +42,15 @@ class yosys_slang(BaseRunner):
         yosys_scr = os.path.join(tmp_dir, "yosys-script")
         mode = params['mode']
 
-        slang_cmd = []
+        slang_cmd = ['-DSYNTHESIS']
+
+        # Ignore content which is unsynthesizable or likely irrelevant for synthesis. The inclusion
+        # of `--ignore-initial` is objectionable as it means the frontend won't pick up memory
+        # initialization.
+        slang_cmd += [
+            '--ignore-timing', '--ignore-initial', '--ignore-assertions'
+        ]
+
         # Some tests expect that all input files will be concatenated into
         # a single compilation unit, so ask slang to do that.
         slang_cmd += ['--single-unit']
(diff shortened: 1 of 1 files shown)


tools/runners/yosys_slang.py
```diff
@@ -26,13 +26,27 @@ def get_mode(self, params):
...
         yosys_scr = os.path.join(tmp_dir, "yosys-script")
         mode = params['mode']
 
-        slang_cmd = []
+        slang_cmd += ['-DSYNTHESIS']
+
+        # Ignore content which is unsynthesizable or likely irrelevant for synthesis. The inclusion
+        # of `--ignore-initial` is objectionable as it means the frontend won't pick up memory
+        # initialization.
+        slang_cmd += ['--ignore-timing', '--ignore-initial', '--ignore-assertions']
```
povik, line 48: It's debatable whether `--ignore-initial` is fair. I'm inclined to push for it since for many applications this is fine, and also the frontend is exposed to many unsynthesizable test cases in this suite (despite the `unsynthesizable` hook from above) so I believe the extent of the frontend's SV support will continue being understated with this flag included. The dashboard results will be more interesting to drive yosys-slang development with these flags in.

povik: I should explain: `--ignore-initial` makes the frontend ignore all content of `initial` blocks

povik:
Gentle ping. If this isn't fixed I'm considering removing yosys-slang from sv-tests altogether. I hope that's understandable.

QuantamHD:
Aren't initial blocks technically synthesis able in FPGA? I'm guessing most of these blocks are in fact simulation, but is it possible to figure out in slang/us if they have any synthesizable structures in them?

povik:
> Aren't initial blocks technically synthesis able in FPGA? 

They are, that's why I think it's worth pointing out.

> but is it possible to figure out in slang/us if they have any synthesizable structures in them?

I'm not sure that can be done robustly so the current plan with yosys-slang is to execute all of initial blocks (unless they are ignored with `--ignore-initial`)

QuantamHD:
Seems reasonable. I can just merge this.

Issue #7105, opened 2025-04-16
opentitan submodule is many years behind


MikePopoloski:
It looks like this hasn't been updated in a very long time. Can this either be fixed up or removed from the test suite?

This is one of the few slang failures but the error slang is complaining about hasn't existed upstream in years.

@tgorochowik

MikePopoloski:
Bump @tgorochowik 

If you don't have the resources to fix this I can put up a PR to remove OpenTitan from the repo. It doesn't make sense to have a many year old version that is simply broken, it's not actually testing anything useful.

tgorochowik:
Sorry for the lack of response earlier. Ideally we'd fix and update this core so it is properly tested. However I agree, given the circumstances and the timeline here, it makes sense to remove it (including all the dedicated dependencies and generator scripts) for now and open a new issue for re-adding it later. Helping with the removal would be very appreciated!

MikePopoloski:
Ok, I'll work on a PR.

tgorochowik:
Removed in #7224

PR #7224, opened 2025-06-14
Remove OpenTitan / earlgrey core

MikePopoloski:
As discussed in #7105


QuantamHD:
@tgorochowik it looks like the self hosted runners are down

tgorochowik:
The CI should be up again

tgorochowik:
Thank you!

PR #7279, opened 2025-07-01
tree-sitter-systemverilog fix attempt

tgorochowik:


tgorochowik:
With the changes in this PR, tree-sitter-systemverilog seems to be working as expected.

However tree-sitter-verilog tries to execute tree-sitter-verilog as well (and fails) - @gmlarumbe shouldn't tree-sitter-verilog be converted to something similar as your systemverilog work? (i.e. tree-sitter parse, but with a different language selected - is that doable?)

gmlarumbe:
Hi @tgorochowik ,

Thanks for taking the time to fix the integration of `tree-sitter-systemverilog`.

Regarding `tree-sitter-verilog` you are right. It was relying previously on Python bindings for tree-sitter but using the `tree-sitter-cli` makes more sense to me. Probably replacing runner `tree_sitter_verilog.py` with something almost similar to `tree_sitter_systemverilog.py` runner would fix it. 

Feel free to change it to see if it works.

On the other hand, the `$HOME/.config/tree-sitter/config.json` adds `$HOME/dev` to the list of parser directories and probably both parsers will be suitable to parse the same code. To select one or the other you can try running the following command for each of the runners:
```
tree-sitter parse --scope "source.verilog" <filename>
tree-sitter parse --scope "source.systemverilog" <filename>
```

Hope this helps.

gmlarumbe:
```python
import os
import shutil

from BaseRunner import BaseRunner


class tree_sitter_verilog(BaseRunner):
    def __init__(self):
        super().__init__(
            "tree-sitter-verilog", "tree-sitter", {"parsing"})

        self.submodule = "third_party/tools/tree-sitter-verilog"
        self.url = f"https://github.com/tree-sitter/tree-sitter-verilog/tree/{self.get_commit()}"

    def prepare_run_cb(self, tmp_dir, params):
        self.cmd = [
            self.executable, 'parse', '--scope', 'source.verilog'
        ]
        self.cmd += params['files']
```

Replacing current `tree_sitter_verilog.py` runner contents with this snippet should fix it.

drom:
@tgorochowik @gmlarumbe 
Why are you hijacking `tree-sitter-verilog` project and replacing it with unrelated `tree-sitter-systemverilog` package?

gmlarumbe:
Hi @drom,

AFAIK `tree-sitter-verilog` is not being replaced. 

The issues/PRs that modify `tree-sitter-verilog` runner have to do with the integration of `tree-sitter-systemverilog` grammar which requires a newer version of `tree-sitter`. Updating tools and the environment to support `tree-sitter-systemverilog` resulted in unwanted errors in `tree-sitter-verilog` and a common approach that shared tools between both grammars was supposed to be the best option (i.e. `tree-sitter-cli`). You can check #6077 if you are curious.

On the other hand I would be more than happy if you decided to take over the update of your grammar here. I was doing it so far since it seemed unmaintained, with many issues and PRs opened and with no answer for more than 2 years.

drom:
I don't see tree-sitter-verilog in the table anymore
image

gmlarumbe:
This probably has something to do with it:

>However tree-sitter-verilog tries to execute tree-sitter-verilog as well (and fails) - [...]

But probably @tgorochowik can tell you better.

kbieganski:
@drom It was a bug. It's been fixed.

PR #7341, opened 2025-07-13
Add ivtest disables

wsnyder:
Blacklist some "new" IV tests that have appeared due to #7338, and test iv-specific versus SystemVerilog-universal stuff.

### DIFF
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -270,6 +270,7 @@ ivtest_file_exclude = [
     'pr2794144',
     # Modules ignored due to module ordering sv-test driver limitation
     'br_gh104b',
+    'check_constant_2',
     'def_nettype_none',
     'implicitport2',
     'implicitport3',
@@ -278,13 +279,42 @@ ivtest_file_exclude = [
     'pr1587634',
     'pr1698659',
     'pr2076425',
+    'pr2792883',
     'shellho1',
     # needs data text file that this test harness doesn't enable finding
     'pr2800985a',
     # primitive table rows are invalid, fails on multiple commercial tools
-    "pr3587570",
+    'pr3587570',
     # Compilation errors with other simulators
-    "pr1520314",
+    'pr1520314',
+    # Synthesis one-hot assertion fails
+    'onehot',
+    # Primitive connect to real
+    'br_gh1182',
+    # Intentional assertion failures
+    'sv_deferred_assert1',
+    'sv_deferred_assert2',
+    'sv_deferred_assume1',
+    'sv_deferred_assume2',
+    'sv_immediate_assert',
+    'sv_immediate_assume',
+    # IEEE 22.2 directives must be on same line
+    'br_gh782b',
+    'br_gh782c',
+    'br_gh782e',
+    'br_gh782f',
+    # Requires iv-specific $abstime/$is_signed
+    'pr2590274a',
+    'pr2590274b',
+    'pr2590274c',
+    'struct_member_signed',
+    'struct_signed',
+    # '%d' with string argument
+    'sv_cast_typedef',
+    # IEEE does not disallow void' of a void function/task, 4 of 5 simulators take it
+    # (Useful in macros that void-call a passed func/task name without knowing dtype)
+    'sv_void_cast_fail1',
+    'sv_void_cast_fail2',
 ]
 
 ivtest_long = ['comp1000', 'comp1001']


kbieganski:
Thanks. CI is complaining about the whitespace at the end of line, once that's fixed I'll merge it

caryr:
I believe some of the items added to the blacklist are actually valid SystemVerilog. Though I will agree they are obscure corner cases. For any invalid tests, these should be changed and iverilog updated to match the standard. For ones that are known to be invalid please file an iverilog report. For example the single line requirement for most compiler directives looks to be an extra limitation added in 1800-2023. Contact me or start a discussion here or in the iverilog repo if you would like to discuss this further.

wsnyder:
@caryr: I agree (I think), the methodology should be people blacklist tests that are wrong, the tool owner is free to file bugs and fix and de-blacklist appropriately.

An alternative where we are required (vs optional after blacklisting) to file bugs on the test providers when a test is broken is not realistically scalable - that would mean I would have had to file literally hundred of bugs by now, I'd just give up on this process as don't have the time if that was to be the requirement.

caryr:
We need to be careful to only blacklist tests that are known/provably to be invalid and not just ones we think are invalid. The goal should be an accurate and valid test suite. Without a standard reference or discussion it tends to fall to a belief not a fact. It also looks like there are also some changes that are needed in how sv-tests determines how to run a specific test to make them SystemVerilog compatible.

For example br_gh1182 is incorrectly labeled in the blacklist and it is marked in our test suite as **invalid Verilog and valid SystemVerilog**. I did some work a while ago to allow Icarus to check all this, but it's likely sv-tests has not been updated to use this extra information to only use the SystemVerilog adapted version/result of the tests.

caryr:
I added the force SystemVerilog flow to specifically make sure that all the Verilog based tests functioned properly when treated as SystemVerilog.

wsnyder:
FWIW I don't know what others do, but unless IEEE indicates something is obviously wrong, I always make sure a test fails on Verilator (which is why I look at it) in addition to at least one other of the big-3 commercial simulators before backlisting.

wsnyder:
>I added the force SystemVerilog flow to specifically make sure that all the Verilog based tests functioned properly when treated as SystemVerilog.

Try making a pull to remove it from the blacklist.  If verilator passes it's probably fine then, if it fails I'd appreciate your looking at why - if you think Verilator's buggy or especially throws an UNSUPPORTED error feel free to still commit it.

wsnyder:
>I added the force SystemVerilog flow to specifically make sure that all the Verilog based tests functioned properly when treated as SystemVerilog.

Try making a pull to remove it from the blacklist.  If verilator passes it's probably fine then, if it fails I'd appreciate your looking at why (based on the output logfile) - if you think Verilator's buggy or especially throws an UNSUPPORTED error feel free to still commit it.

caryr:
I've not looked at it in detail, so it's pure speculation that the regression files Icarus is supplying are just being loaded in sv-tests and the override behavior between the files is being ignored. It could be the same reason Icarus has failing tests when run in the sv-tests suite, but passes 100% locally.

I also try to check things I'm not 100% sure on using the Big-three simulators as a reference, but I have seen cases where all three had related bugs that were all different, so sometimes it's just a reading of the standard and the general trend of the three.

caryr:
Continuing to use br_gh1182 as an example. It is marked as a CE (Compile Error) in the vlg (Verilog) file and then marked normal in the fsv (Force SystemVerilog) file. The intention is when SystemVerilog is forced the original definition regarding how to run and the results expected in the vlg file will be overridden by the fsv contents.

MikePopoloski:
FWIW slang in its default mode is quite strict with language conformance and fully supports all language features up through the latest version of the LRM, so if slang results match Verilator I think there's a good chance it's the test itself that is busted.

wsnyder:
Looking at slang on iv_tests it does suggest:

1. some more disables are needed (e.g `$abstime` test disables).  I @MikePopoloski maybe you could look at those you think that slang is correct about, and make a blacklisting pull?  Otherwise for those that fail slang and Verilator, I'll eventually notice, and make a blacklist pull.
2. As @caryr pointed out, the ivtest lists are not properly being used - we should be running assuming only latest-standard SV behavior.
3. there's cases that slang can't get right because they require simulation runtime or otherwise.  I'm assuming @MikePopoloski  or otherwise will mark those as parser-only tests, since I selfishly won't notice as track only through the complete simulation runtime (verilator vs icarus differences).

Also it would make sense that icarus should be passing it's own test list, but is not, most of those seem to be related to bucket 2 (correction) above.

Anyhow above items are future improvements and shouldn't block this pull which reduces case 1 items at least.

caryr:
@MikePopoloski and I spent some time a few years ago getting ivtest mostly clean with slang. What was left were actual limitations in slang. That work was the incentive for me to add the force SystemVerilog flow to make sure things stayed clean. Some new test have switch to a new flow Steve is working on and it does not currently support forcing SystemVerilog and may not currently be identified by sv-tests as tests to run.

For Icarus it looks like there are two major buckets. The tests that are failing because the SystemVerilog overrides are not being applied correctly in sv-tests and for VCD output, etc. The tests assume there is a _work_ sub-directory where output files are generated. I didn't check everything, but the ones I did look at fall into these categories.

I checked the slang results and some should be blacklisted, some fall into the categories we have been discussing and a few look like issues in the slang parser. For example the standard in 23.2.2.4 says:
> Defaults can be specified only for input ports and only in ANSI style declarations. A default shall not be
> specified for a port declared as interconnect (an interconnect port).

MikePopoloski:
I will do a pass and add more disables after this PR lands, or fix slang if I find legit issues. I didn't see any tests that require actual simulation in order to get the right result btw.

@caryr I couldn't find the test you were referring to with the snippet about ports but I'll look again. I agree that you should try to get sv-test to know the correct mode / expected failure state for the SystemVerilog specific tests. Can you provide a PR?

@tgorochowik @kbieganski it would be nice to get more hands available that can help speed up landing of PRs like this. Perhaps @wsnyder and I could be added as members to help out?

caryr:
module_inout_port_list_def.v - inout ports cannot have a default value

caryr:
I'll try to get something that fixes the test list generation and that should clean up most/all of the issues for Icarus along with improving the results for the other tools.

wsnyder:
I updated the comment for sv_foreach6.v.  I agree the grammar allows `[]` however this is a no-operation and almost certainly not what any user would code without it being a mistake. If icarus wants to continue to treat it as a NOP without warning, as it does now, I have no objection.  However, for Verilator I want to be able to fail a compile when a user does this, similar to 4 other simulators failing it, so still believe it should be blacklisted.

MikePopoloski:
Seems like a weird line to draw to me -- it's nice to warn or even error by default on bugprone code but as you say it's legal, so the tool should have some mode to allow it. It's not true to say the commercial tools all fail on it either; Xcelium and RivieraPro both support it, for example. It's not like there aren't thousands of other ways to write code that does nothing useful but Verilator will still happily compile either. Shrug.

caryr:
Since it's valid SystemVerilog it should be allowed. This is a test suite to check SystemVerilog, not just the parts we agree with. Individual tools can choose to do things differently, but that is a non-conformance to the standard and should be flagged as an expected fail for that tool with an appropriate explanation.

caryr:
I'm fine blacklisting code that uses tool specific constructs, but we should not be blacklisting "Perfectly valid SystemVerilog"

kbieganski:
> it would be nice to get more hands available that can help speed up landing of PRs like this. 

Unless I'm missing something, I don't think this PR is suffering from any hold-ups on our side. I'm waiting until a consensus is reached on which tests should be blacklisted.

wsnyder:
>I'm fine blacklisting code that uses tool specific constructs, but we should not be blacklisting "Perfectly valid SystemVerilog"

I don't consider this "perfectly valid" as I think of that bar being real code like a user would write, but anyhow accept if we want the test suite to cover everything parseable.

Pushed removing the foreach blacklist.

@kbieganski with that am not aware of other objections, please merge.

kbieganski (approved):

PR #7342, opened 2025-07-14
Add github workflow_dispatch trigger

wsnyder:
This adds a button in the github UI under each action workflow to allow it to be run when pressed.


kbieganski:
What's the use case for this?

wsnyder:
I can run it on my fork when I want without a fake push.  it's a pretty standard practice to have this.

kbieganski (approved):

PR #7366, opened 2025-07-18
Use verilator itself instead of --top-module

wsnyder:
Rather than determining the top module via the runner python, rely on Verilator's knowledge.

This may allow additional tests that don't have the modules in the proper order to now pass, but we'll see what the fallout is, so draft only for the moment.


wsnyder:
@kbieganski this is ready to merge. Verilator has more failures in total but these are real.  Other tool count changes are noise. Thanks.

PR #7372, opened 2025-07-19
Add more ivtest exclusions

MikePopoloski:
Reasons for the exclusions are documented in the comments. This addresses almost all of the remaining slang failures. I believe the remaining ones are actual bugs in slang that I will fix.

@caryr @wsnyder for review.

### DIFF
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -221,9 +221,14 @@ ivtest_file_exclude = [
     # though the non-blocking assignment will have no effect.
     'program3b',
     # A wire with 'real' type is not allowed (commercial tools disallow)
+    'br_gh156',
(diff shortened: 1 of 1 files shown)


wsnyder:
See also #7371, @kbieganski if want do this one, then I will clean #7371

wsnyder:
I've added these to my pull request #7371, I suggest reviewing that and close this.

I object to removing macro_comment_multiline, so commented it out in #7371.  While I agree IEEE says this is not allowed, this is a very common thing for people to do, and every simulator I tried supports this (one warns and works, rest do not even warn).

MikePopoloski:
Ok, I will close this and move comments over there.

PR #7433, opened 2025-08-02
A few more ivtest and basejump excludes

MikePopoloski:
Reasons for the two ivtest excludes are documented in the file.

The basejump exclude is due to calling a task from within a final block, which is disallowed by the LRM. Some tools error and some just warn on this.

@wsnyder @caryr

### DIFF
--- a/conf/generators/meta-path/basejump.json
+++ b/conf/generators/meta-path/basejump.json
@@ -60,7 +60,8 @@
 		"bsg_test_BaseJumpSTL_bsg_nonsynth_axi_mem.sv",
 		"bsg_wormhole_router_pkg.sv",
 		"test_bsg_clock_params.sv",
-		"bsg_dmc_xilinx_ui_trace_replay.sv"
+		"bsg_dmc_xilinx_ui_trace_replay.sv",
+		"bsg_nonsynth_profiler.sv"
 	],
 	"fake_topmodule": true,
 	"results_group": "imported"
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -369,6 +369,12 @@ ivtest_file_exclude = [
     'task_nonansi_vec2',
     # Test should expect fail: IEEE does not allow .* to connect to empty ports
     'br_gh530',
+    # The LRM does not seem to support the idea of leaking enum members out of a surrounding
+    # struct scope, though some tools support it. VCS does not.
+    'enum_in_struct',
+    # All tools allow this. The LRM contains some conflicting wording on this, but shows
+    # an example that implies it should be allowed.
+    'sv_export_fail1',
 ]
 
 ivtest_long = ['comp1000', 'comp1001']


wsnyder (commented):
Seems reasonable.

While there can we remove 'par_mismatch'?  It prints FAILED on all simulators including icarus, I'm not sure what it was originally trying to test.

MikePopoloski:
par_mismatch seems legit to me. slang says:
```
../../../third_party/tools/icarus/ivtest/ivltests/par_mismatch.v:21:9: error: parallel specify path source type 'logic' has different width than destination type 'logic[7:0]'
    (in => out) = 2;
     ~~ ^~ ~~~
```

The LRM has a whole section on parallel paths, 30.4.5:
```
The operator => shall be used to establish a parallel connection between source and destination. In a parallel
connection, each bit in the source shall connect to one corresponding bit in the destination. Parallel module
paths can be created only between sources and destinations that contain the same number of bits.

Parallel connections are more restrictive than full connections. They only connect one source to one
destination, where each signal contains the same number of bits. Therefore, a parallel connection may only
be used to describe a module path between two vectors of the same size. Because scalars are 1-bit wide,
either *> or => may be used to set up bit-to-bit connections between two scalars.
```

MikePopoloski:
Probably the simulators are just ignoring the specify blocks since they don't affect simulation?

caryr:
The par_mismatch test is marked as a compile error and why is documented in the file. I agree with Mike it is valid though maybe the -gspecify is messing up sv-tests. Icarus by default does not have specify block functionality enabled.

wsnyder:
Just checking you agree it shouldn't fail, which its current printing FAIL does, so it should be excluded?

caryr:
It's printing FAILED since the parallel connection is invalid and this should be caught during elaboration.

caryr:
> Probably the simulators are just ignoring the specify blocks since they don't affect simulation?

They can impact simulation delays irrespective of using a SDF file.

caryr:
The par_mismatch test should fail elaboration (a compilation error) since the width of the parallel connection source and destination do not match. If for some reason this is compiled without error it displays FAILED to indicate that something was missed in the elaboration stage.

wsnyder:
@caryr but this test does elaborate and prints "FAILED" on every simulator I tired **including icarus** (see https://chipsalliance.github.io/sv-tests-results/?v=icarus+ivtest+regress-vlg_par_mismatch_iv).

caryr:
To get the compile error message in Icarus you have to enable specify blocks (**-gspecify**):
```
iverilog -gspecify iverilog/ivtest/ivltests/par_mismatch.v
iverilog/ivtest/ivltests/par_mismatch.v:21: error: For a parallel connection the source/destination width must match found (1/8).
1 error(s) during elaboration.
```
And it looks like `slang` is properly detecting this error based on what @MikePopoloski posted.

caryr:
It looks like **-gspecify** needs to be added to the `iverilog` runner. Without this the entire specify block is ignored.

wsnyder:
Ok, if the runner is fixed I withdraw the exclusion request.

caryr:
Since we have determined that par_mismatch is valid Verilog code and that both slang and Icarus when given the correct option will report this failure at elaboration time. Why is your opinion of it based on updating the Icarus runner to properly add the `-gspecify` option when needed?

This started as a request to remove a couple tests which are completely unrelated to this second issue. I propose this pull request get merged and if you want, create an issue that the Icarus runner and possibly other tools need to have their runners updated. It's slightly more complicated from the Icarus side since we also have tests that specifically disable the specify block support so all that would need to be worked out like was done when all the tests were forced to be run in SystemVerilog mode.

wsnyder:
Agree this is ok to merge.  My point was intended to concur with not excluding it.

PR #7522, opened 2025-08-28
Tests: Disable Ivtests br_gh1175f

wsnyder:
Disables br_gh1175f, which redefines a UDP by the same name.

Behavior of multiple definitions of primitives/modules is outside IEEE and depends on library handling.

Icarus and one closed-source simulator error (which this test required).
Three closed-source simulators warned (false-failing this test as return good edit status).
Two closed-source simulators silently accepted (false-failing this test as return good edit status).

### DIFF
--- a/generators/ivtest
+++ b/generators/ivtest
@@ -375,6 +375,9 @@ ivtest_file_exclude = [
     # All tools allow this. The LRM contains some conflicting wording on this, but shows
     # an example that implies it should be allowed.
     'sv_export_fail1',
+    # Redefining the same UDP is above what IEEE specifies; it's an error on some tools,
+    # warning on others (which pick first or last defined), and ignored on yet others.
+    'br_gh1175f',
 ]
 
 ivtest_long = ['comp1000', 'comp1001']


wsnyder:
@caryr @kbieganski

wsnyder:
@kbieganski can you please merge?

caryr:
Sorry for not replying earlier. Go ahead and remove this for now, but I think this is covered by a compilation namespace requirement. When I have time I'll look at everything that has been removed and either fix Icarus or give a good reason why this should be allowed. I agree there are some places where multiple alternatives are allowed and going to the lowest common denominator seems wrong so I think sv-tests needs to be updated to delineate a given tool does not support the given functionality as the tests expects.

kbieganski:
I was waiting for some more discussion, as I don't have much of an opinion on this.  I'll merge it for now.

> I think sv-tests needs to be updated to delineate a given tool does not support the given functionality as the tests expects.

Suggestions on how to do this are welcome. I don't want this to water down the report, or suddenly have tons of exceptions per tool, so we'll have to be careful with this.

PR #7588, opened 2025-09-14
Fix delayed assignment to class member

wsnyder:
1. Fixes a delayed assignment to a class member. IEEE requires such assignments to be static.
2. Blacklist hdlconvertor_std2017_p753.  Code is illegal; simple_bus is referenced but not defined

All simulators agree with this.  Note Icarus does not support these tests for other reasons, but if I comment out the offending other code it agrees.


wsnyder:
@kbieganski think is good to go. The new failures are random noise and unrelated. Verilator didn't get a pass with the change due to a general timeout issue I'll resolve separately.

PR #7589, opened 2025-09-14
For Verilator, use lower -j due to UVM tests etc

wsnyder:
The UVM tests are timing out, this sets up less top parallelism to leave more CPUs free for Verilator to parallelize its build on.

Draft as might need tweaking or a straight timeout increase.


wsnyder:
@kbieganski is ready for review.  I made several passes and seems stable.

I had to make two changes, first for Verilator have fewer parallel jobs so that cores are free for Verilator parallelism, and also increase timeouts because sometimes multiple large UVM jobs land at same time.

PR #7604, opened 2025-09-17
Delete uvm-fail tests

wsnyder:
This test was marked simulator only, but can't pass on simulators.

1. The 16.12--property-uvm-fail.sv  interface was misconnected, which this would fix; other tests disabled have similar or other issues.

```
diff --git a/tests/chapter-16/16.12--property-uvm-fail.sv b/tests/chapter-16/16.12--property-uvm-fail.sv
--- a/tests/chapter-16/16.12--property-uvm-fail.sv
+++ b/tests/chapter-16/16.12--property-uvm-fail.sv
@@ -97,7 +97,7 @@ endclass
 module top();
     env environment;

-    mem_ctrl_if dif();
+    mem_ctrl_if dif(.clk(dif.clk), .read(dif.read), .write(dif.write), .addr(dif.addr), .dout(dif.dout), .din(dif.din));

     mem_ctrl dut(.clk(dif.clk), .read(dif.read), .write(dif.write), .addr(dif.addr), .dout(dif.dout), .din(dif.din));
```

2. With that the code, 16.12--property-uvm-fail.sv correctly on Verilator gets a UVM_FAIL, but there's no way to know if the UVM_FAIL should return bad exit status or not (most simulators appear not to), so AFAIK the test can't be known to pass or not.

There are similar issues on all these tests, I don't believe they are written well enough to be part of the tests.  If you want to instead invest to fix them to pass on multiple simulators, that's fine too of course.  Thanks.


wsnyder:
@kbieganski

wsnyder:
@kbieganski think is ok to go, note tree sitter failures are unrelated.

PR #7614, opened 2025-09-20
Fix more UVM timeouts

wsnyder:
Missed two timeout increases.  uvm_agent_active.sv will pass with this.  uvm_files already passed but wanted to match other timeouts.

@kbieganski thanks!


wsnyder:
Found a bunch more needing fixing.

@kbieganski good to go.  The failures are expected-fail tests that actually pass once the timeout was fixed and will be addressed in another pull as they don't pass on other simulators too.  Also really timeout should have never allowed an expected-fail test to pass, something to fix later.

wsnyder:
To avoid merge conflict reverting tests also in (#7604).  Still think good to go.

PR #8334, opened 2026-04-01
Work around duplicate modules in ibex; align circt-verilog to Slang

fabianschuiki:
Add the `-Wno-duplicate-definition` flag to the circt-verilog and Slang runners for builds of the Ibex core. The source code contains two implementations of the primitive cells, one generic and one Xilinx-specific, which provide duplicate definitions. This should probably be fixed in the Ibex generator, too, but I have very little visibility into how that all works.

Also copy a few more command-line flags from Slang over to circt-verilog to align the latter more closely with the former.

### DIFF
--- a/tools/runners/Slang.py
+++ b/tools/runners/Slang.py
@@ -59,10 +59,13 @@ class Slang(BaseRunner):
 
         tags = params["tags"]
 
-        # The Ariane core has syntax errors with stream concat operators and duplicate definitions.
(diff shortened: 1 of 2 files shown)


tobiasgrosser:
Nice. Thank you, @fabianschuiki.

tobiasgrosser:
@kbieganski, this seems to be ready-to-merge.

tobiasgrosser:
@kbieganski, pushing this up in the todo list post easter.

PR #8403, opened 2026-04-22
Forward slang args to slang

jpienaar:
This argument doesn't exist on circt-verilog, just pipe it through to slang.


tobiasgrosser (approved):
Thank you, @jpienaar!

jpienaar:
Neither the extra passing or extra failing makes sense, this affects only ariana ones and they were failing due to incorrect flag before already.

PR #8465, opened 2026-05-06
[IBEX][FuseSoC] Fix IBEX fusesoc config

Scheremo:
First time contributor, please let me know if I missed any contribution guidelines or similar - happy to invest reasonable effort to adjust this PR.

As explained in [the lowRisc documentation](https://opentitan.org/book/hw/ip/prim/index.html#mappings), the `lowrisc:prim` library is virtual, and should be mapped via the CLI. Currently, no mapping is passed, and all potential `lowrisc:prim` candidates are added to the elaboration flow; this causes warnings in `verilator`, and errors in `circt-verilog`, since it leads to module definition duplication.

This PR fixes this issue by choosing the de-facto correct mapping of `prim_generic`, since `prim_xilinx` would require external dependencies that are currently not mapped.

### DIFF
--- a/conf/fusesoc-configs/ibex-sim.yml
+++ b/conf/fusesoc-configs/ibex-sim.yml
@@ -12,7 +12,7 @@ description: Full ibex core test
 top_module: ibex_simple_system
 tags: ibex
 path: third_party/cores/ibex
-command: fusesoc --cores-root third_party/cores/ibex run --target=sim --setup lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast
+command: fusesoc --cores-root third_party/cores/ibex run --target=sim --setup --mapping lowrisc:prim_generic:all:0.1 lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast
 conf_file: build/lowrisc_ibex_ibex_simple_system_0/sim-verilator/lowrisc_ibex_ibex_simple_system_0.vc
 test_file: ibex-sim.sv
 timeout: 100
--- a/conf/requirements.txt
+++ b/conf/requirements.txt
@@ -9,7 +9,7 @@ make_var
 markupsafe
 pytablewriter
 GitPython
-simplesat==0.8.2
-git+https://github.com/lowRISC/edalize.git@ot
-git+https://github.com/lowRISC/fusesoc.git@ot
+simplesat==0.9.1
+edalize==0.6.1
+fusesoc==2.4.3
 orderedmultidict
--- a/conf/fusesoc-configs/veer-eh1-sim.yml
+++ b/conf/fusesoc-configs/veer-eh1-sim.yml
@@ -13,7 +13,7 @@ top_module: tb_top
 tags: veer-eh1
 path: third_party/cores/veer-eh1
 command: fusesoc --cores-root third_party/cores/veer-eh1 run --target=sim --setup --build-root build/veer-eh1_sim chipsalliance.org:cores:VeeR_EH1:1.8
-conf_file: build/veer-eh1_sim/sim-verilator/chipsalliance.org_cores_VeeR_EH1_1.8.vc
+conf_file: build/veer-eh1_sim/chipsalliance.org_cores_VeeR_EH1_1.8/sim-verilator/chipsalliance.org_cores_VeeR_EH1_1.8.vc
 test_file: veer-eh1-sim.sv
 timeout: 180
 compatible-runners: verilator slang circt-verilog
--- a/conf/fusesoc-configs/veer-eh1-synth.yml
+++ b/conf/fusesoc-configs/veer-eh1-synth.yml
@@ -13,7 +13,7 @@ top_module: veer-eh1_wrapper
 tags: veer-eh1
 path: third_party/cores/veer-eh1
 command: fusesoc --cores-root third_party/cores/veer-eh1 run --target=synth --setup --build-root build/veer-eh1_synth chipsalliance.org:cores:VeeR_EH1:1.8
-conf_file: build/veer-eh1_synth/synth-vivado/chipsalliance.org_cores_VeeR_EH1_1.8.tcl
+conf_file: build/veer-eh1_synth/chipsalliance.org_cores_VeeR_EH1_1.8/synth-vivado/chipsalliance.org_cores_VeeR_EH1_1.8.tcl
 test_file: veer-eh1-synth.sv
 timeout: 180
 compatible-runners: yosys-synlig yosys yosys-sv zachjs-sv2v icarus moore moore-parse odin sv-parser tree-sitter-verilog verible verible_extractor Surelog slang-parse circt-verilog


tobiasgrosser:
Thank you @Scheremo!

kbieganski:
CI failing on fusesoc:
```
15:23:52 | Traceback (most recent call last):
15:23:52 |   File "/root/sv-tests/sv-tests/./generators/fusesoc", line 93, in <module>
15:23:52 |     with open(core_conf_path, "r") as f:
15:23:52 | FileNotFoundError: [Errno 2] No such file or directory: '/root/sv-tests/sv-tests/build/lowrisc_ibex_ibex_simple_system_0/sim-verilator/lowrisc_ibex_ibex_simple_system_0.vc'
```
Seems like `conf_file` is incorrect? Not super familiar with fusesoc

VecoMr:
In my view, the current issue is that we are using a version of fusesoc that does not support the `--mapping` option. As a result, the generator runs the command and fails silently. This mean the file is not created which is why we are getting this error.
After updating fusesoc to version >= 2.4.3 we must also update the `conf_file` path in `veer-eh1-sim.yml` and `veer-eh1-synth.yml`, because the output directory for the build now defaults to build/$VLNV for the `--build-root` option.

Scheremo:
> In my view, the current issue is that we are using a version of fusesoc that does not support the `--mapping` option. As a result, the generator runs the command and fails silently. This mean the file is not created which is why we are getting this error. After updating fusesoc to version >= 2.4.3 we must also update the `conf_file` path in `veer-eh1-sim.yml` and `veer-eh1-synth.yml`, because the output directory for the build now defaults to build/$VLNV for the `--build-root` option.

Right. I'm happy to implement this here, if this is appropriate to have in one PR.

kbieganski:
Yes, feel free to do so

Scheremo:
> Yes, feel free to do so

The latest commit should set the correct fusesoc version and updates the veer core configs. Can we give it another go with the CI?

tobiasgrosser:
Amazing, @kbieganski, we may need a rerun of the CI.

conf/fusesoc-configs/ibex-sim.yml
```diff
@@ -12,7 +12,7 @@ description: Full ibex core test
 top_module: ibex_simple_system
 tags: ibex
 path: third_party/cores/ibex
-command: fusesoc --cores-root third_party/cores/ibex run --target=sim --setup lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast
+command: fusesoc --cores-root third_party/cores/ibex run --target=sim --setup --mapping lowrisc:prim_generic:all:0.1 lowrisc:ibex:ibex_simple_system --RV32E=0 --RV32M=ibex_pkg::RV32MFast
```
marnovandermaas, line 15: Do you know why you need to add this mapping value? I thought the generic mapping was the default. The upstream simple system does not say to add this and I have built this with Verilator: https://github.com/lowRISC/ibex/tree/master/examples/simple_system#building-simulation
What kind of warnings are you seeing in Verilator without the mapping?

conf/requirements.txt
```diff
@@ -11,5 +11,5 @@ pytablewriter
 GitPython
 simplesat==0.8.2
 git+https://github.com/lowRISC/edalize.git@ot
```
marnovandermaas: Is this dependency still necessary?

marnovandermaas (commented):
Just a few thoughts on my end but don't let these block you.

conf/fusesoc-configs/ibex-sim.yml
Scheremo: So the error pattern is that without the `mapping`, for some reason, fusesoc picks **both** prim_generic and prim_xilinx. You can actually see this e.g. on the verilator runs, it generates a bunch of `%Warning-MODDUP` to indicate duplicated modules, since both `prim_generic` and `prim_xilinx` are imported. 
Verilator will still go ahead, and by lucky coincidence might use the `prim_generic` implementation. Classic case of two unreasonable default behaviours kind of cancelling out, I guess.
Slang/CIRCT however is a bit stricter in how module duplication is handled, and this causes an error.

conf/requirements.txt
Scheremo: I personally don't know, I tend to avoid touching anything using `fusesoc` / `edalize`. 
Maybe someone with a fusesoc background can comment?

conf/fusesoc-configs/ibex-sim.yml
Scheremo: FWIW the standalone `Slang` target only passes because it's called with `-Wno-duplicate-definition --ignore-unknown-modules`. Without those `Slang` also complains that the tech cells of `prim_xilinx` are undefined, and that both `prim_generic` and `prim_xilinx` are imported.

Scheremo:
I do not currently have rights to merge this myself; if anyone could push that button for me I would highly appreciate it!

conf/fusesoc-configs/ibex-sim.yml
marnovandermaas: I did this build in the upstream Ibex repository and I don't see these MODDUP warnings. Can you try whether you see those warnings there as well?

Scheremo: I think what you're asking is outside of the scope of this PR. I'm not trying to make any claims about upstream Ibex, so I am hesitant to invest time to reproduce a workflow that may or may not yield different results.

You can verify that what I say is correct by checking the [sv-tests history scoreboard](https://chipsalliance.github.io/sv-tests-results/?v=verilator+ibex+ibex), navigate to Verilator / Ibex and check the tool output. You'll find both the `prim_xilinx` cells included as well as the `prim_generic` cells, and you'll see the MODDUP warnings. I'm attaching a screenshot of the message here that contains all elements. You can also run the updated version of sv-tests (in this PR) and see that those warnings no longer occur. 

image

conf/requirements.txt
kbieganski: This causes the CI to fail unfortunately:

```
ERROR: Cannot install -r conf/requirements.txt (line 14) and edalize 0.3.3 (from git+https://github.com/lowRISC/edalize.git@ot) because these package versions have conflicting dependencies.
```

Scheremo: Thanks for rerunning the CI! I'll have a look what the most straight-forward fix for this would be.

Scheremo: @kbieganski simplest solutions seems to be to pin newer versions of `edalize` and `simplesat`. Can we rerun the CI?

Scheremo:
Fixed the edalize version conflict; can we do another CI run? @kbieganski

kbieganski:
Seems like the same warning is still happening in Verilator.

Scheremo:
> Seems like the same warning is still happening in Verilator.

Can you point me to where you see this? In the report I downloaded from CI the `prim_xilinx` entries are gone, the `MODDUP` warnings are gone but the error for verilator remains. Attached the log I'm looking at.

[ibex-sim.sv.log](https://github.com/user-attachments/files/28643529/ibex-sim.sv.log)

kbieganski:
I meant the error (thought it was a warning that can be disabled, but it's not).

That error should probably be fixed in Ibex (see https://github.com/verilator/verilator/issues/7352)

kbieganski (approved):

kbieganski:
Thanks @Scheremo!

PR #8604, opened 2026-06-05
Add z3 for Verilator tests

wsnyder:
Recently added uvm_ac_config_db has constraints and so Verilator needs the z3 solver package.


wsnyder:
@kbieganski  if you can please merge, thanks

PR #8628, opened 2026-06-11
Add an Arcilator runner column

AmurG:
# Add an Arcilator runner column

This PR adds a runner for [arcilator](https://github.com/llvm/circt/tree/main/tools/arcilator),
CIRCT's simulation backend, alongside the existing `circt-verilog` column.
While the `circt-verilog` column measures how much SystemVerilog the CIRCT
frontend accepts, the new `Arcilator` column measures how much of it survives
the full lowering pipeline and, for simulation rows, actually executes.

The whole column is 237 added lines: a single 217-line runner plus the CI
matrix entry, the `arcilator` make target, and two fusesoc
compatible-runners lines.

## What the column means

* **Simulation rows** (`:type: simulation`): a two-stage script (so the log
  shows each stage's exact command): `circt-verilog -o design.mlir` followed
  by `arcilator design.mlir --run` (JIT execution). A green cell means the
  design compiled, lowered, and ran, and nothing in the run log contradicted
  the exit status (see the demote-only check below).
* **simulation_without_run / elaboration rows** (including the SoC cores
  generated from `third_party/cores`): `circt-verilog -o design.mlir`
  followed by `arcilator design.mlir --disable-output`. This is an
  *elaboration contract*: the full arcilator lowering pipeline runs, but the
  design is not executed. A green cell means "arcilator can build this", not
  "this was simulated".
* **parsing / preprocessing rows**: frontend only (`--parse-only` / `-E`),
  same as the `circt-verilog` runner.

Frontend options mirror `tools/runners/circt_verilog.py` so the two
CIRCT-backed columns agree on what the frontend accepts.

## Top selection and anti-false-green handling

For simulation rows the runner honors an explicit `:top_module:` and
otherwise accepts only the unique uninstantiated module root; with zero or
several candidates it refuses to pick an arbitrary entry point and exits
with 126. Choosing a wrong top could simulate something other than what the
test intends and report a pass for it. Frontend-only modes pass no `--top`
in the ambiguous case and let slang elaborate all roots.

A status-0 simulation whose own log prints `UVM_FATAL`, a nonzero
`UVM_ERROR`, or `TEST FAILED` is demoted to a failure. These checks only
ever demote; nothing promotes a failing run to a pass. They are disabled on
`should_fail` rows, where a forced nonzero status would be published as the
expected failure, so a demotion could manufacture a green cell for code the
toolchain wrongly accepted. For the same reason the top-selection refusal
exits in the crash range (126) instead of 1, so it can never satisfy a
`should_fail` row either. UVM rows run under a fixed `ulimit -v` so a
runaway JIT simulation fails deterministically instead of starving the CI
host.

## Provenance and build

Both tools are built from the existing `third_party/tools/circt-verilog`
submodule (official `llvm/circt`, pinned by dependabot); no new submodule
and no fork pin is introduced. The `arcilator` make target reuses the build
directory configured by the `circt-verilog` target and just installs the
additional `arcilator` binary. The CI job shares the `circt-verilog` ccache
namespace, making the duplicate build nearly free after the first run, and
caps run parallelism because JIT simulations can be memory-hungry. The CI
entry uses `runners_filter: Arcilator` so the job does not duplicate the
`circt-verilog` column.

## Current expectations (honest baseline)

One structural caveat at the current submodule pin: `arcilator --run`
cannot yet execute the designs `circt-verilog` emits (JIT entry-point
synthesis is part of the CIRCT simulation work that is being upstreamed
separately), so positive simulation rows are currently red after compiling
and lowering; simulation greens at this pin come only from `should_fail`
rows. The column exists precisely to track that simulation slice turning
green as CIRCT main improves.

Against upstream CIRCT main, expect:

* most LRM chapter simulation rows that the frontend accepts to compile and
  lower, turning green as arcilator's JIT execution path lands;
* most UVM rows (`tests/uvm`, `uvm-examples`, easyUVM) to be **red**: the
  Moore/LLHD lowering work needed to elaborate and run UVM class code is
  being upstreamed to llvm/circt separately, and this column is intended to
  track that progress honestly. Rows that need DPI C sources, dual
  `hdl_`/`hvl_` tops, or `:simvariants:` execution are red or refused in
  this column; that functionality was removed from this PR to keep it small
  and reviewable, and can be proposed separately later if wanted;
* SoC core rows to be green only up to the elaboration contract, and only
  where the frontend already accepts them.

The column will improve as CIRCT main does; the runner deliberately
contains no workarounds that would make it look better than the toolchain
is.

Note: upstream `arcilator` currently has no step-budget flag, so runaway
simulations are bounded by the per-test `:timeout:` (and the fixed
`ulimit -v` for UVM rows) rather than a JIT step limit.

## Validation

Run locally with `circt-verilog` and `arcilator` on `PATH`:

```sh
export OUT_DIR=$PWD/out CONF_DIR=$PWD/conf TESTS_DIR=$PWD/tests \
       RUNNERS_DIR=$PWD/tools/runners THIRD_PARTY_DIR=$PWD/third_party
# simulation row exercising the :assert: convention
./tools/runner --runner Arcilator -t chapter-20/20.5--real-bits-conv.sv \
    -o $OUT_DIR/logs/Arcilator/20.5--real-bits-conv.sv.log
# should_fail rows (simulation-mode and elaboration-mode)
./tools/runner --runner Arcilator -t chapter-6/6.5--variable_multiple_assignments.sv \
    -o $OUT_DIR/logs/Arcilator/6.5.log
./tools/runner --runner Arcilator -t chapter-5/5.6--wrong-identifiers.sv \
    -o $OUT_DIR/logs/Arcilator/5.6.log
# version reporting
./tools/runner --runner Arcilator --version -o $OUT_DIR/logs/Arcilator/version
```

`make format` is clean for the added files.

## What this PR does NOT claim

* It does not claim UVM simulation support in upstream CIRCT today; the UVM
  rows are expected to be red until the corresponding CIRCT work lands.
* A green elaboration/SoC cell does not claim the design was simulated —
  only that the arcilator lowering pipeline accepted it.
* It does not modify any test, generator, or third-party pin, and it does
  not change the behavior or results of any existing runner column. The only
  conf change adds `Arcilator` to the compatible-runners of the two fusesoc
  simulation rows, mirroring how `circt-verilog` was added, so the column
  does not silently lack those rows.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tools/runners/Arcilator.py
+++ b/tools/runners/Arcilator.py
@@ -1,0 +1,214 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2026 The SymbiFlow Authors.
+#
+# Use of this source code is governed by a ISC-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/ISC
+#
+# SPDX-License-Identifier: ISC
+
+import os
+import re
+import shlex
+import shutil
+
+from BaseRunner import BaseRunner
+
+# Demote-only: a status-0 simulation printing one of these failed itself.
+SELF_REPORTED_FAILURE_RES = (
+    re.compile(r"^UVM_FATAL\b(?!\s*:\s*0\b)"),
+    re.compile(r"^UVM_ERROR\b(?!\s*:\s*0\b)"),
+    re.compile(r"\bTEST\s+FAILED\b", re.IGNORECASE),
+)
+
+
+class Arcilator(BaseRunner):
+    """Run tests through the CIRCT Verilog frontend and arcilator.
+
+    Simulation rows are compiled ahead of time (``arcilator --emit-llvm``,
+    ``llc``, link against CIRCT's shipped arc runtime) and run to
+    quiescence by the generated ``<top>_main`` driver. Unsupported
+    constructs are rejected while compiling, so this column can score
+    fewer green cells than an in-process ``--run`` column. Other rows
+    stop after ``arcilator --disable-output``.
+    """
+    def __init__(self):
+        super().__init__(
+            "arcilator",
+            executable="arcilator",
+            supported_features={
+                "preprocessing", "parsing", "elaboration", "simulation",
+                "simulation_without_run"
+            })
+        self.submodule = "third_party/tools/circt-verilog"
+        self.url = f"https://github.com/llvm/circt/tree/{self.get_commit()}"
+
+    def can_run(self):
+        return all(map(shutil.which, ("circt-verilog", "arcilator", "llc")))
+
+    def run_subprocess(self, tmp_dir, params):
+        output, rc = super().run_subprocess(tmp_dir, params)
+        should_fail = params.get("should_fail") == "1"
+        if rc == 71 and should_fail:
+            return (
+                output + "[arcilator] timed out; a timeout is not a "
+                "toolchain rejection and must not satisfy should_fail\n", 126)
+        # Never demote should_fail rows; that would create a green cell.
+        if rc != 0 or should_fail:
+            return output, rc
+        if params["mode"] == "simulation":
+            if any(pattern.search(line)
+                   for line in output.splitlines()
+                   for pattern in SELF_REPORTED_FAILURE_RES):
+                return output, 1
+            # A run printing zero ':assert:' lines passes the harness check
+            # vacuously; demote when the sources print them unconditionally.
+            if ":assert:" not in output and any(
+                    '$display(":assert:' in text
+                    for text in self._source_texts(params)):
+                return (
+                    output + "[arcilator] exited 0 without printing the "
+                    "test's ':assert:' verdict lines; demoting\n", 1)
+        return output, rc
+
+    def _source_texts(self, params):
+        for path in params["files"]:
+            try:
+                with open(path, encoding="utf-8", errors="ignore") as f:
+                    yield f.read()
+            except OSError:
+                continue
+
+    def _unique_module_root(self, params):
+        """Return the sole uninstantiated module, or None if ambiguous."""
+        texts = list(self._source_texts(params))
+        module_def_re = re.compile(
+            r"\bmodule\s+([A-Za-z_][A-Za-z0-9_$]*)"
+            r"\s*(?:#\s*\([^;]*?\)\s*)?[(;]", re.DOTALL)
+        modules = []
+        for text in texts:
+            for name in module_def_re.findall(text):
+                if name not in modules:
+                    modules.append(name)
+        roots = [
+            name for name in modules if not any(
+                re.search(
+                    r"(?<!\bmodule\s)\b" + re.escape(name) +
+                    r"\s*(?:#\s*\([^;]*?\)\s*)?[A-Za-z_][A-Za-z0-9_$]*\s*\(",
+                    text, re.DOTALL) for text in texts)
+        ]
+        return roots[0] if len(roots) == 1 else None
+
+    def _uses_uvm_runtime(self, params):
+        return "uvm" in params["tags"] or any(
+            "run_test" in text for text in self._source_texts(params))
+
+    def _write_simulation_stages(self, f, tmp_dir, params, top):
+        # <top>_main is arcilator's generated driver; it returns at
+        # quiescence, once no model-side wakeup remains pending.
+        with open(os.path.join(tmp_dir, "driver.cpp"), "w") as drv:
+            drv.write(
+                f"extern \"C\" void {top}_main(void);\n"
+                f"int main(void) {{ {top}_main(); return 0; }}\n")
+        libdir = os.path.join(
+            os.path.dirname(os.path.dirname(shutil.which("llc"))), "lib")
+        cxx = "clang++" if shutil.which("clang++") else "c++"
+
+        f.write("arcilator design.mlir --emit-llvm -o design.ll || exit $?\n")
+        # PIC code, since the driver links into a default (PIE) executable.
+        f.write(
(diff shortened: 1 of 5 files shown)


AmurG:
@fabianschuiki any ideas on who would be the right person on chipsalliance side for a review of this column ?

tools/runners/Arcilator.py
kbieganski, line 1: Does this runner have to be so big? Most runners are far below 100LoC. Perhaps some of this functionality should be in Arcilator itself?

AmurG:
Sorry about the claudeslop — TLDR: one real piece (top-module detection) belongs in CIRCT; opened llvm/circt#10680 for it, linked both ways. I dug into the rest, and most of it honestly stays runner-side:

- **Top detection + wrapper synthesis** (~130 lines of source regex) → circt-verilog behind an opt-in `--detect-top` flag (default unchanged): llvm/circt#10680.
- **DPI** stays — arcilator is a no-compiler MLIR JIT (vs Verilator's source→C++→cc), so the clang build can't move in without giving it a compiler dependency; it already takes the prebuilt `.so` via `--shared-libs`.
- **The `CIRCTSupport.py` grading** stays — it's UVM/sv-tests verdict semantics (UVM_ERROR alongside a `$finish` exit-0, the `:assert:` logparser, TEST-FAILED markers) on top of exit codes the tools already set right; a frontend/JIT can't know UVM semantics.

So it comes down by the ~130 top-detection lines to ~300, not Verilator's ~100 — the rest is genuine two-tool-JIT + UVM-grading harness logic. Getting to ~100 would need arcilator to grow a combined `arcilator file.sv --run` front-door — bigger change; say the word if that's the direction you want.

tools/runners/Arcilator.py
```diff
@@ -0,0 +1,437 @@
...
+                        "printf '%s\\n' " + shlex.quote(
+                            "[Arcilator] C/C++ header inputs are available "
+                            "through include directories and are not passed to "
+                            "circt-verilog") + "\n")
+            if wrapper_roots:
+                f.write(
+                    "printf '%s\\n' " + shlex.quote(
+                        "[Arcilator] simulation test has multiple portless "
+                        "module-root tops; using a generated wrapper top that "
+                        "instantiates: " + ", ".join(wrapper_roots)) + "\n")
```
kbieganski: Are these generated prints necessary?

kbieganski:
Any reduction of the diff is welcome.

AmurG:
Done — slimmed the column to the minimal honest baseline: **237 added lines across 5 files, was 767 across 6** (`Arcilator.py` is now 217 lines and the separate `CIRCTSupport.py` grading library is gone entirely). Force-pushed as a clean 3-commit series on current master; PR description updated to match. As before, this rework is Claude-assisted and I reviewed it.

- The runner mirrors `circt_verilog.py`'s frontend handling and runs a two-stage script: `circt-verilog -o design.mlir`, then `arcilator design.mlir --run` (`--disable-output` for elaboration rows).
- Top selection: explicit `:top_module:` or the unique uninstantiated module root, otherwise a simulation row is refused with exit 126 (crash range, so a refusal can never satisfy a `should_fail` row). The regex top-guessing heuristics, the generated dual-top wrapper, and the `$printtimescale` special case are gone. On the circt side, the `--detect-top` offload (llvm/circt#10680) was closed, so the runner keeps this one minimal check.
- Grading shrank to a demote-only hook inlined in the runner: a status-0 simulation printing `UVM_FATAL`, a nonzero `UVM_ERROR`, or `TEST FAILED` is demoted; nothing ever promotes, and `should_fail` rows are never demoted. Longer term I agree the right fix is in arcilator itself (nonzero exit on `$fatal`/failed assertions), which would delete most of even this.
- Dropped from the PR: the DPI C-source sidecar and `:simvariants:` execution. Disclosed cost: the handful of imported UVM-example rows that need those go red/refused in this column for now. The removed pieces are preserved out-of-tree and can be proposed separately later if you want them.

Validated locally: the slim runner and the previous one produce identical verdicts on a 19-row sample (LRM simulation rows, `should_fail` rows, explicit-`:top_module:` rows, UVM rows, parsing/preprocessing rows, and the multi-root refusal case), and `make format` is clean. Today's dashboard should not move at all: at the current pin positive simulation rows are red anyway (no JIT entry point yet), so the drops only become visible as red/refused UVM-example rows once arcilator execution lands. Ready for another look.

AmurG:
PR updated, rebased on master. Let me know if this can merge.

Dashboards generated with this runner: https://amurg.github.io/sv-tests-results-v1/ and https://amurg.github.io/sv-tests-results-v2/

AmurG:
@kbieganski try now looks small enough

tools/runners/Arcilator.py
```diff
@@ -0,0 +1,214 @@
...
+        # loud crash-range exit can never satisfy a should_fail row.
+        budget = max(5, int(params["timeout"]) - 5)
+        f.write(
+            f"timeout -k 5 {budget} ./sim\nrc=$?\n"
+            "if [ $rc -eq 124 ] || [ $rc -eq 137 ]; then\n")
+        f.write(
+            "    printf '%s\\n' " + shlex.quote(
+                f"[arcilator] simulation exceeded its {budget}s wall-clock "
+                "budget; treating this as a runner failure, not a verdict") +
+            "\n    exit 126\nfi\nexit $rc\n")
```
kbieganski, line 141: Can we use templates for these rather than just writing strings to a file? Some tests and other files are generated from templates.

kbieganski (changes requested):
Thanks for reducing the diff, looks pretty good. Got one comment, after than and if CI passes, I'll merge it.

PR #8640, opened 2026-06-15
Add liblz4 for Verilator

wsnyder:
Ibex on verilator requires FST tracing and so liblz4.


wsnyder:
@kbieganski good to merge - the non-verilator changes are false noise.

Issue #8609, opened 2026-06-07
Fix uvm-examples to properly compile


wsnyder:
@kdanecki  you recently added uvm-examples (@kbieganski note too). There are several problems with the sv-tests setup that are not Verilator bugs.  When adding new tests it would be appreciated if before adding you look at the message all tools print to see if it indicates a setup problem or a real tool bug.

First, the constraints, #8604 will fix that, please merge that.

Then the following tests do not seem to have all of the files, please either fix these to get verilator (etc) to pass, or disable them, or file a bug on Verilator. At first look none of these seem Verilator bugs or I wouldn't ask ;) - though the last two I'm less sure of..  Thanks

```
uart_example,False,1,,360,%Error: third_party/tests/uvm-examples/uart_example/agents/modem_agent/modem_agent_pkg.sv:25:10: Cannot find include file: 'modem_config.svh'
uvm_tb_build_ss_tb,False,1,,360,%Error: third_party/tests/uvm-examples/uvm_tb_build_ss_tb/sub_system_tbs/pss_tb/tb/apb_prober.sv:36:8: Can't find definition of scope/variable: 'APB'
uvm_virt_seq_find_all,False,1,,360,%Error: third_party/tests/uvm-examples/uvm_virt_seq_find_all/simple_ovc/simple_sequencer.svh:23:5: Define or directive not defined: '`uvm_sequencer_utils'
biquad_example,False,1,,360,%Error: third_party/tests/uvm-examples/biquad_example/tb/agents/signal_agent/signal_seq_item.svh:80:20: Member 'sprint' not found in class 'm_uvm_printer_knobs'
uvm_use_models_pipelined_item_done,False,1,,360,%Error: third_party/tests/uvm-examples/uvm_use_models_pipelined_item_done/pipelined_item_done/mbus_pipelined_agent/mbus_seq_item.svh:105:22: Member 'sprint' not found in class 'm_uvm_printer_knobs'
```

There are also these timeouts.  These might be verilator bugs but please check against a commercial tool and make sure pass there.  It might be they just need the timeout raised too.

```
uvm_generation_seq_persistence,False,1,,360,%Error-DIDNOTCONVERGE: third_party/tests/uvm-examples/uvm_generation_seq_persistence/./common/top_tb.sv:21: Inactive region did not converge after '--c\
onverge-limit' of 10000 tries
uvm_id_register,False,71,,360,Timeout: > 360s
[...]

PR #8655, opened 2026-06-17
Disable some uvm-examples

wsnyder:
Disables tests mentioned in #8609.  Really these should get more attention as presumably they were added for some effect, but for now just remove.


wsnyder:
@kbieganski think good to go. The tree sitter failures are random/unrelated.

PR #8681, opened 2026-06-22
build(deps): bump cmake version to >= 3.28 for circt-verilog

VecoMr:
Circt-verilog vendors `slang`as a build dependency. Recent `slang` releases raised their `cmake_minimum_required` to `3.28`, while the CI container  (`ubuntu:jammy-20221130`) only ships `CMake 3.22.1` through apt. This makes the `circt-verilog` build fall during configuration, as example we have this [PR](https://github.com/chipsalliance/sv-tests/pull/8677) where the CI failed.

This PR installs a recent CMake via pip specifically for for the `circt-verilog` job.

This is my first contribution to this project, so feel free to point out anything I got wrong or could do differently.

### DIFF
--- a/.github/workflows/sv-tests-ci.yml
+++ b/.github/workflows/sv-tests-ci.yml
@@ -54,7 +54,8 @@ jobs:
             deps: bison clang tcl-dev flex libfl-dev pkg-config libreadline-dev cmake pkg-config
             submodules: third_party/slang
           - name: circt-verilog
-            deps: cmake clang ninja-build lld
+            deps: clang ninja-build lld
+            cmake_ver: ">=3.28"
             submodules: llvm
     env:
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
Nice. Thank you @VecoMr!

VecoMr:
I find it strange that my changes are affecting the tests, because the new step I added is ignored by everyone except `circt-verilog`.

tobiasgrosser:
This is indeed surprising. On the positive side, `circt-verilog` seems to work again. So this PR seems to be ready-to-merge, but we should keep this surprising test noise in mind. Maybe someone else has seen this before?

fabianschuiki (approved):
LGTM! Thanks for fixing this 👍

PR #8927, opened 2026-08-15
Fix nondeterministic simulation test oracles

dsellerbrock:
## Summary

Fix six simulation tests whose oracles depended on invalid file access, scheduler races, host-language parsing, or simulator-specific exit-code behavior.

## Root causes and fixes

- Reopen the self-created `$ungetc` file in read mode before using input functions; the prior test used a write-only stream.
- Move the minimal event trigger to the Inactive region so its waiter is deterministically armed first; same-Active-region trigger/wait ordering is a race.
- Compare the tagged-union active value in SystemVerilog instead of asking Python `eval` to parse the assignment-pattern text emitted by `%p`.
- Use `$fatal(1, ...)` for three `:should_fail_because:` SVA runtime tests. `$error` reports a runtime error but does not portably require a nonzero host-process exit status.

## Validation

- All three positive simulations compile, run, and pass `tools/logparser.py` with the current Icarus SystemVerilog implementation.
- All three negative SVA simulations compile, reach the intended assertion failure, and exit 1.
- Slang 11.0.415 accepts all six sources with 0 errors and 0 warnings.
- Every command ran with a 45-second per-process CPU cap and 1 GiB aggregate RSS cap.

These changes correct the test evidence; they do not weaken or special-case simulator behavior.


caryr:
This PR fixes #9024

